PR #8, opened 2018-02-20
feat(grid): expose objectIdProperty with bindable prop

jmzagorski:
This pull requests adds a bindable property for the `objectIdProperty` parameter on the `setItems` dataview method. Many of my APIs had to use a property other than "id" so slickgrid throws an error `Each data element must implement a unique 'id' property`. Adding this bindable property allows the consumer to specify their unique key if it is other than "id" and defaults to "id" if not specified.


ghiscoding:
Could you tell what does this bring? I always used a `gridId` but never used a `datasetId`, what is that for?

jmzagorski:
The `gridId` is used by slickgrid to identifier the container the grid will be rendered in:

``this.grid = new Slick.Grid(`#${this.gridId}`, this.dataview, this.columnDefinitions, this._gridOptions);``

The `datasetId` is for the unique id property in your json dataset that the dataview needs. In your step-by-step wiki you have a json dataset with properties: 

```javascript
{
        id: i,
        title: 'Task ' + i,
        duration: Math.round(Math.random() * 100) + '',
        percentComplete: randomPercent,
        start: `${randomMonth}/${randomDay}/${randomYear}`,
        finish: `${randomMonth}/${randomDay}/${randomYear}`,
        effortDriven: (i % 5 === 0)
};
```

If you change the name of the `id` property to `foobarId`, you will receive an error from the dataview as mention in my first comment. By adding the `datasetId` we can default to using the name `id`, which is the default behavior of slickgrid, but now the consumer of the custom element can specify a custom property name like `foobarId` that represents their unique key in their json dataset.

ghiscoding:
Oh wow I learned something new today about SlickGrid, that is great, I remember having issues in the past and having to add an extra "id" all lower case to fix that. Ahh SlickGrid, you always surprise me lol

jmzagorski:
I actually missed another `setItems` call in `refreshGridData`. Will update pull request

ghiscoding:
Take your time, I actually won't release another version until I include the "Export to File" feature, in the weekend or early next week. Just advise me when your PR are ready to be merged.

ghiscoding:
Taking another look at your EDIT with the example and I would rather have this option as a `gridOption`. I personally prefer to have all the options defined in the ViewModel rather than the View itself since the theory is to have less logic in the View. We could leave it as `bindable` option, but I would also like to have it as a `gridOption`. Would it be possible to update your PR with that in mind? The other advantage with the `gridOption` is that you can define some option globally for your project, see the [Wiki - Global Options](https://github.com/ghiscoding/aurelia-slickgrid/wiki/Global-Options)

To do the change that I'm asking for, you would have to do these steps
1. add a new option in the [gridOption.interface](https://github.com/ghiscoding/aurelia-slickgrid/blob/master/aurelia-slickgrid/src/aurelia-slickgrid/models/gridOption.interface.ts)
2. set a default global option as "id" in the [globalGridOption.ts](https://github.com/ghiscoding/aurelia-slickgrid/blob/master/aurelia-slickgrid/src/aurelia-slickgrid/global-grid-options.ts) file.
3. if you keep the `bindable` option, you would have to merge your option with the `gridOption` of the user which always have higher priority.

jmzagorski:
This PR should be ready whenever you are...no rush.

ghiscoding:
How about `datasetIdPropertyName` which might be long, it could also be `idPropertyName` which is shorter and near what you saw in the `DataView`

jmzagorski:
Went with the longer name. This PR is ready to merge.

ghiscoding:
Looks good thanks, I assume you tested it locally?
I'll probably merge it tomorrow, need a break today.

jmzagorski:
Yes. I've been testing it by building the native-modules and coping that folder over to my actual project. Hopefully I can start adding some unit tests once I finish the grid on my project.

And no rush on getting these in, especially since I plan on submitting another small pull request and opening and a few issues just so the discussion can start.

ghiscoding:
Yeah that is also how I test all the `client` samples. Inside my Build NPM Script, at the end of the Build it copies the dist to all client node_modules folders. 

One more thing, in your future PR, could you add comments in this form `/** what it's for */` for the interface property names.
I found very recently that adding comment in that form, actually shows up in the intelisense of VSCode, so I'm slowly adding them. My next PR will fill all the gridOptions properties.

Example of the `/** comment */`, hovering over Filter `type` property gives me this (see comment in grey at the bottom)
tooltip

Out of subject, I merged the PR for the new feature of SlickGrid - Row Detail Panel. 
Right after, Ben (6pac) released a new [version 2.3.13](https://github.com/6pac/SlickGrid/releases/tag/2.3.13) there's a few fix, which you can see on the [link](https://github.com/6pac/SlickGrid/releases/tag/2.3.13), that I'm happy to see, and particularly the show/hide flags for ColumnPicker/GridMenu of the last 2 checkboxes (ForceFitColumn & SyncResize). So my next version will bump SlickGrid in my `package.json`

jmzagorski:
hmm that's strange I thought I added a comment to the `datasetIdPropertyName` in the interface file as you had on some other properties  with `/** */`. I was switching branches and stashing often so I must have overlooked that when finalizing the commit. I will make sure to do that in the future.

ghiscoding:
You can add the comment on your new variable, I will do all the other variables. I am actually in this process right now with my Angular-Slickgrid repo, so I will just have to copy+paste to Aurelia later. It's long to do on my side but will pay off later by helping development :)

Thanks

ghiscoding:
Just made a quick test with some of my examples and they all throw errors because you made the new property a required field. 
All options of the `GridOption` interface should be optional else we won't even be able to even declare a grid. I know that at the end, I will nonetheless always have certain values (with the Global Option file) which are required for the grid to work. 

Unless you have better suggestions to deal with this, I think we will have to continue with all properties being optional. Thoughts?

jmzagorski:
You are correct, it is suppose to be optional. I have not coded in typescript for a few years so I easily forget these extra steps to be explicit with the types. I'll make a note that all grid options should be optional (as it is in slickgrid's api). I think this is a good example of where we can implement unit tests in the future to document what should done. A test such as "all grid options should be optional", seems pretty silly, but it would have helped with this issue.

As for adding `this.dataview.setItems(this._dataset, this._gridOptions.datasetIdPropertyName || 'id');`, I am not sure it is needed because of this code `    this.gridOptions = { ...GlobalGridOptions, ...binding.gridOptions };` setting the default value to `id`. I think the library should be explicit and throw the error if the consumer sets `datasetIdPropertyName: null` in their view model because they are not allowed to explicitly do that in slickgrid.

If you agree with that I will update this pull request with a question mark in the interface:

```
  /** Unique property name on the dataset used by Slick.Data.DataView */
  datasetIdPropertyName?: string;
```

ghiscoding:
If SlickGrid throws the error itself and the TypeScript Build works then I agree. There's a lot of these condition that I add in the code because of the TS Build.

jmzagorski:
Okay I've pushed that update. Let me know if I need to do anything else.

ghiscoding:
Merged

Next version release will be after I copy over the "Export to File" feature that I've done in Angular this week. Possibly in the weekend or early next week. Got a couple of minor fixes in the queue also. 

Thanks

PR #11, opened 2018-02-22
fix(grid): init dataset on attached if it has data

jmzagorski:
This pull requests ensures that the `AureliaSlickgridCustomElement` will use the `dataset` when it is created without relying on `datasetChanged`

**Current State**
The `_dataset` is set in two methods: `attached` and `datasetChanged`. However, what happens if the view model is created and the dataset already has the data locally in a service, cache etc? That would mean that `datasetChanged` would never fire and the grid will always be empty. I experienced this problem when navigating back and forth between view-models. The first time I navigate to the view model, the grid is populated because I fetch the data from an API. However, when I navigate away and then back to that view model the grid becomes empty because `dataset` already has the data when the custom element is reattached to the DOM. 

**Future State**
The simple change in this pull request will ensure the `_dataset` is initially set to the dataset in case the `datasetChanged` method is never fired:


ghiscoding:
Good catch and really great to see all your explanation. Thanks

PR #12, opened 2018-02-23
fix(events): unsubscribe from all events on detached

jmzagorski:
Use the Slick.EventHandler for subscriptions so we can use `unsubscribeAll`. I think with the original code we would have had to track each subscribed function and unsubscribe individually.

Also, as a precaution, add hasOwnProperty check on property enumeration


ghiscoding:
Do you have a way to test it?

jmzagorski:
I tested it by including a breakpoint in chrome at the `unsubscribeAll`. I saw the `_eventHandler` had 39 handlers and it iterated through all of them to unsubscribe and and empty the array.

I would be more than happy to add a unit test, but I was not sure how quickly you wanted this out the door.

ghiscoding:
If you have time to start with a simple unit test that would be awesome. Technically a new version won't happen until the coming days, so you probably have time. When I said test, I meant some kind of manual test to confirm that it did get called and you just confirmed it. Thanks

Issue #9, opened 2018-02-22
Creating a filter factory to create new filters in the filter.service


jmzagorski:
I was reviewing your prior commits and saw you improved the filtering by adding a [filter interface](https://github.com/ghiscoding/aurelia-slickgrid/blob/master/aurelia-slickgrid/src/aurelia-slickgrid/models/filter.interface.ts). I think we can take this one step further by using a factory to create the filters and allow Aurelia to inject any number of dependencies that are needed in the filters. I'll explain with current vs future state below:

Current State:
In the [filter interface](https://github.com/ghiscoding/aurelia-slickgrid/blob/master/aurelia-slickgrid/src/aurelia-slickgrid/models/filter.interface.ts) you manually instantiate a filter using a `switch(filterType)` and passing in the i18n service that is injected into the `FilterService`.

Future State:
1. We add a property on the `Filter` interface that requires every Filter implementation to have a `FilterType` value (ie the `SelectFilter` would have `filterType = FilterType.select`. 
2. We code a new `FilterFactory` such as:

```typescript
@inject(All.of('GridFilters'))
export class FilterFactory {

constructor(private filters: Filter[]) {}

public createFilter: Filter (filterType: string) {
  return this.filters.find(f => f.filterType === filterType);
  // or throw an exception if a filter is not found
}
```
3. Tell Aurelia about our new objects in `main.ts`:

```
aurelia.container.registerSingleton('GridFilters', InputFilter);
aurelia.container.registerSingleton('GridFilters', SelectFilter);
aurelia.container.registerSingleton('GridFilters', SingleSelectFilter);
aurelia.container.registerSingleton('GridFilters', MultipleSelectFilter);
```

Benefits: 
1. This allows Aurelia to handle the dependency graphs for each filter. This means we can inject any number of dependencies into the filters without bloating the `FilterService`. This is a huge benefit. One example I can think of is using the `BindingEngine` from `aurelia-framework` to watch the `collection` property on the select list filters ( :-) could be a future pull request). Right now the library assumes this `collection` will have all the data when the view model is ready. However, what if the consumer is using a [generator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Iterators_and_Generators) instead of a promise, or they are fetching the data at some later time after the view model is already created or what if the consumer deletes a record in the grid that was in the select filter and should no longer be in the select filter (sorry for going off on tangent, but I have thought about this for a few days)

[...]

PR #13, opened 2018-02-23
refactor(Filter): use a FilterFactory for Filter instantiation

jmzagorski:
This is related to issue https://github.com/ghiscoding/aurelia-slickgrid/issues/9, which has all the details. I was going to include a test, but getting that setup was taking longer than expected and I didn't want this branch to hang out too long without feedback. Catching up on the unit tests will probably be in a separate pull request unless you would want those in this request.

I did run a test using your examples with `npm run start:dev` and all your filters seemed to work and were created via this new `FilterFactory`

### DIFF
--- a/aurelia-slickgrid.wiki/Custom-Filter.md
+++ b/aurelia-slickgrid.wiki/Custom-Filter.md
@@ -1,0 +1,100 @@
+### Description
+You can also create your own Custom Filter with any html/css you want and/or jQuery library you wish to use. Aurelia template (View) are not supported at this point, if you wish to contribute on that end then I certainly accept PR (Pull Request). 
+
+#### Limitations
+- as mentioned in the description, only html/css and/or jQuery libraries are supported.
+  - this mainly mean that Aurelia templates (Views) are not supported (feel free to contribute).
+- SlickGrid uses `table-cell` as CSS for it to display a consistent height for each rows (this keeps the same row height/line-height to always be the same). 
+  - all this to say that you might be in a situation were your filter shows in the back of the grid. The best approach to overcome this is to use a modal if you can or if the library support `append to body container`. For example, you can see that `multiple-select.js` support a `container` and is needed for the filter to work as can be seen in the [multipleSelectFilter.ts](https://github.com/ghiscoding/aurelia-slickgrid/blob/master/aurelia-slickgrid/src/aurelia-slickgrid/filters/multipleSelectFilter.ts#L26)
+
+### Demo
+[Demo Page](https://ghiscoding.github.io/aurelia-slickgrid/#/slickgrid/example4) / [Demo Client Component](https://github.com/ghiscoding/aurelia-slickgrid/blob/master/aurelia-slickgrid/src/examples/slickgrid/example4.ts) / [Custom InputFilter.ts](https://github.com/ghiscoding/aurelia-slickgrid/blob/master/aurelia-slickgrid/src/examples/slickgrid/custom-inputFilter.ts)
+
+### How to use Custom Filter?
+1. You first need to create a `class` using the [Filter interface](https://github.com/ghiscoding/aurelia-slickgrid/blob/master/aurelia-slickgrid/src/aurelia-slickgrid/models/filter.interface.ts). Make sure to create all necessary public properties and functions. 
+ - You can see a demo with a [custom-inputFilter.ts](https://github.com/ghiscoding/aurelia-slickgrid/blob/master/aurelia-slickgrid/src/examples/slickgrid/custom-inputFilter.ts) that is used in the [demo - example 4](https://ghiscoding.github.io/aurelia-slickgrid/#/slickgrid/example4)
+2. There are two Simply set the filter type to `FilterType.custom` and instantiate your class with `new` (can also pass DI in the constructor if you wish). Here is an example with a custom input filter: 
+```javascript 
+// define you columns, in this demo Effort Driven will use a Select Filter
+this.columnDefinitions = [      
+  { id: 'title', name: 'Title', field: 'title' },
+  { id: 'description', name: 'Description', field: 'description', type: FieldType.string,
+    filterable: true, 
+    filter: {
+       type: FilterType.custom,
+       customFilter: new CustomInputFilter() // create a new instance to make each Filter independent from each other
+    }
+  }
+];
+
+// you also need to enable the filters in the Grid Options
+this.gridOptions = {
+   enableFiltering: true
+};
+```
+
+### Default Search Term(s)
+If you want to load the grid with certain default filter(s), you can use the following optional properties:
+- `searchTerm` (single value) with `singleSelect` or `select`
+- `searchTerms` (array of values) with `multipleSelect`
+
+For example, setting a default value into an `input` element, you can simply get the search term with `columnDef.filter.searchTerm` and set the default value in jquery with `$(filterElm).val(this.searchTerm);`
+
+### Collection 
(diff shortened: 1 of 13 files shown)


ghiscoding:
I assume it might also affect the way I call the `customFilter` in my example 4, meaning this [line](https://github.com/ghiscoding/aurelia-slickgrid/blob/master/aurelia-slickgrid/src/examples/slickgrid/example4.ts#L58)?

This will interfere slightly with some of my current feature on the Angular side (preset a grid with certain filters/orders/pagination which we need at work) but it doesn't seem to be too bad. If you want a previous of the preset feature, here's the first big commit of the past 2 days (https://github.com/ghiscoding/Angular-Slickgrid/commit/6dc8875f224c24b59ea3ccf5082801feddef1222)

I'll look at your code possibly tomorrow, and might merge it then. I'll also be starting to copy over the other feature that I've done in Angular at the beginning of the week which is Export to File (csv/text).

jmzagorski:
I tried to not make this effect custom filters except for the fact that a custom filter class needs the `filterType` property. I thought about making this an optional property on the `Filter` interface, but didn't go through with it. We can make it optional if it makes this more flexible, and it wont break your current Filter interface API. The `FilterFactory` is really only for predefined filters created in your filter.service so your `CustomInputFilter` [here](https://github.com/ghiscoding/aurelia-slickgrid/blob/master/aurelia-slickgrid/src/examples/slickgrid/example4.ts#L58) can be called the same way

So are filter presets saved by the users for when they return to the page? Let me know what files(s) and method(s) this will interfere with on your Angular side. Maybe if I know the conflicts I can apply some different techniques to make it work better in Aurelia. You can wait to merge this until we investigate more.

ghiscoding:
The example with the customFilter is just a week old and I would be very very surprised if anyone used it yet. So I don't mind making changes if need be. Whatever you think is the best approach with less coding to do in creating a custom filter.

jmzagorski:
That property makes sense for any filter that gets put into the `GridFilters` plugin, but for custom filters outside of the library, users will most likely have an empty property or use the `FilterType.custom` enum and inject the filter into their view model or new it up as you did in your example. In this case, that property does not make sense. 

I still debate which is better and I don't have a lot of typescript experience to know which option is better to take. Either way it should be affect functionality, just the interface. What do you think?

ghiscoding:
Not sure which property you're talking about but for the `FilterType.custom` yes it was designed for that purpose, to be used with a Custom Filter. I'm ok with calling custom filter the way it is currently in the example by newing it up. I don't have enough practice with the Factory Pattern, I understand how it works but not enough knowledge to comment on the structure and implementation of it. I'm also ok with changing the interface if it doesn't affect too many of it's properties, and again the fact that these Filter and Custom Filter are so new (1 week old), it's still ok to modify it slightly if need be.

I did a checkout locally of your PR and it all works as expected and don't see any issues. We still new up the custom filter and that's ok. I'll leave it up to you to decide if you want to do any changes, just let me know when it's ready to merge. Thanks

jmzagorski:
I thought about this a little more and made the property optional. I also allowed it to be a string so it makes more sense with custom filters, which adds another optional for users to register their custom filters. Instead of explaining here I added my notes to the `Custom-Filter.md` wiki. I first committed that wiki page in its original form and then added my changes so you could more easily see my changes. Let me know what you think. If you approve and it works on your end then this pull request is done.

Also, i tend to commit a lot even if it is small so 1. my changes are saved and 2. you can see my changes in logical groups. Feel free to squash all my commits when doing a pull request if you do not want your history cluttered.

ghiscoding:
Awesome thanks, I like the fact that you took the time to do 2 commits, that really helps to see the differences. I forgot to move the `wiki` folder into the `doc` folder, we can use the `doc` folder in future PR. I also like that you moved the singleton registration in the `index.ts`, it makes sense to see all of what's happening directly in there. 

Maybe unrelated but just so you know, since I saw you added the type to `FilterType | string`. Let say you want to permit the users to pass a string but only certain strings, you can define a type like this
```javascript
export type SortDirectionString = 'asc' | 'ASC' | 'desc' | 'DESC';
```
Then instead of type `string`, I would call the type `SortDirectionString`, so they can still type a string but if they go outside of these strings, TypeScript will tell the user and fail. Not sure if you knew this or not, since you said TypeScript is in the back of your head, and it's probably unrelated to yours changes but anyhow it's nice to know about. 

I use the "Squash & Merge" button in Github, but for some reason when I pull the commits with VSCode, it puts back all the commits, not sure why lol. But it's ok, I don't mind at all to see all these commits, it helps me revised the changes and create my release notes. 

Are you done with the commits/PR?

jmzagorski:
Thanks for that typescript info, I did not know that. I plan to move my huge app to typescript so any insight such as that helps!

If you are satisfied with the new changes and it works on your end then yes this PR is done.

ghiscoding:
Thanks I will merge by the end of the day, so you still have few hours in case you missed something.

I'm starting to use also a lot more of the ENUM type with string and the code I shown you above, so that I can save to LocaleStorage or DB in some cases. The code about the sort direction string is because of that I reason, I want to save the presets (filters/sorters) into local storage and if the ENUM is used without string, I would be saving numbers which isn't very helpful. 

So these pieces of codes help for saving
```javascript
export type SortDirectionString = 'asc' | 'ASC' | 'desc' | 'DESC';

export enum SortDirection {
  asc = 'asc',
  ASC = 'ASC',
  desc = 'desc',
  DESC = 'DESC'
}
``` 
When I mix the 2 together, in the code then we can use the Types OR input the string. Like this
```javascript
// so the type of `direction` is 
// direction: SortDirection | SortDirectionString;
filters: [
{ columnId: 'firstName', direction: SortDirection.ASC }
// or type a string
{ columnId: 'lastName', direction: 'DESC' }
```
When saved in LocalStorage, they both will have the string (not a number) which is what I wanted.

I just learned that this week, so I'm happy to share :)

jmzagorski:
I like that idea. I use LocalStorage as well for my older slickgrid application to save the columns they show vs hide with the column picker because there are probably 50 or so columns they can choose from

ghiscoding:
@jmzagorski 
I replied to your post in a separate issue #14 to avoid touching too many subjects and for work in progress. I'd like to have your input in that other post. Thanks

ghiscoding:
@jmzagorski 
I forgot to check the CLI version, and another reason why I have that CLI is because that one is more tricky compare to the others. What I found recently is that CLI (RequireJS) doesn't seem to like import from something like `from './models;'`, what it wants to see is `from './models/index;'` (basically just index to the import.

The kind of errors that shows up in the CLI start/build is something like the following
```javascript
...
Tracing examples/slickgrid/example1...
Tracing examples/slickgrid/example10...

 ------- File not found or not accessible ------
| Location: C:/OwnSyncDev/GitHub/aurelia-slickgrid/client-cli/node_modules/aurelia-slickgrid/dist/amd/filters.js
| Requested by: C:\OwnSyncDev\GitHub\aurelia-slickgrid\client-cli\src\examples\slickgrid\example10.js
| Is this a package? Make sure that it is configured in aurelia.json and that it is not a Node.js package
 -----------------------------------------------
```

It took me a while to figure out that it was only missing the `index` in the import. Webpack doesn't care if you put it or not but RequireJS and maybe SystemJS seems to care. I only found that last week in my previous release, I didn't update them all yet, but I'll do that today.

jmzagorski:
I did not know that, wonder why the CLI doesn't support that. Sorry about changing those and thanks for fixing it.

Issue #20, opened 2018-03-01
Filters are out of sync since the Filter Factory implementation


ghiscoding:
**I'm submitting a bug report**
* **Browser:**
all
* **Language:**
TypeScript

**Current behavior:**
The current `Filter` class has multiple properties like `columnDef`, `callback`, `searchTerm`, ... all of which are associated to the current column filter element. However the Filter Factory has a `this.filters.find` function which has a side effect of overwriting (or re-using) previous column filter. For example, let say column 1 and 4 are input filter, column 2 and 3 are select filter, then the current behavior of that column 1 will create a filter in the factory because it didn't find any but then when it comes back to the factory for column 4, it finds a filter and returns it but a new `columnDef` associated and column1 filter becomes out of sync (now it's columnDef is associated to column 4, so when I start typing in column 1 it tries to filter data against column 4.

**Expected/desired behavior:**

* **What is the expected behavior?**

* **What is the motivation / use case for changing the behavior?**
I think this is a bad design from my side and `Filter` should be separated into 2 different classes, making something like `ColumnFilter` and `ElementFilter` classes were `ElementFilter` should only have the behavior of a DOM element (that one would stay with the Filter Factory) but the ColumnFilter class would have each `columnDef` and other related properties. This new issue was not there before the Filter Factory was introduced. I don't see this problem in Angular since I haven't put the Factory pattern in that repo yet and the previous code was creating a new instance on every filter element which is a difference compare to the Filter Factory which does a `find()`.

For example, the first print screen below shows the `columnDef` before creating the filter and we can see it's associated to "percentComplete" but then it goes in the createFilter Factory, it finds a filterType and returns it but then that filter already has a `columnDef` associated to it, which is "Title". So now, when I start typing a filter in "percentComplete", it's actually trying to filter on "title"

Before creating the filter, it's `columnDef` is "percentComplete"
![filter1](https://user-images.githubusercontent.com/643976/36826786-c37e6d08-1cdd-11e8-81e7-28b31fa4d1f6.png)

After creating the filter, it's now associated to `columnDef` of "title"
![filter2](https://user-images.githubusercontent.com/643976/36826797-d0922980-1cdd-11e8-9374-da1ed77ba47b.png)

jmzagorski:
Oh yes you are right, the factory class should create a new instance every time it is called. I did not test this thoroughly enough since I only tested it on one column, sorry.

[...]

PR #21, opened 2018-03-01
 fix(filterFactory): create a new filter instance on every call 

jmzagorski:
This pull request fixes singleton filters in the `FilterFactory` because `All.of` was used in the `FilterFactory` constructor. It should close https://github.com/ghiscoding/aurelia-slickgrid/issues/20. 

**Changes**
1. Filter registration was changed to `registerTransient` instead of `registerSingleton`
2. The `FilterFactory` uses the Aurelia's `container.getAll` method to find the correct filter every time `FilterFactory.createFilter` is called

I'd like to mention one of several other alternatives. I could use the `filterType` value as the registration key. That would change the developer registration to `aurelia.container.registerTransient(FilterType.input, Filter.inputFilter)` instead of `aurelia.container.registerTransient(FILTER_PLUGIN_NAME, Filter.inputType). `

**Alternative Benefits**
1. Do not need a `FILTER_GRID_PLUGIN` constant 
2. Do not need to call `container.getAll` to iterate over all the filters. We can call `container.get(filterType)` instead.

**Alternative Cons**
1. Registration key could have conflicts unless we created a `FILTER_GRID_PLUGIN_PREFIX` constant so the registration would look like `aurelia.container.registerTransient(FILTER_GRID_PLUGIN_PREFIX + FilterType.input, Filter.inputFilter)` and we would get the filter by `container.get(FILTER_GRID_PLUGIN_PREFIX + filterType)`.
2. The filters would not be registered together.

However, I went with the current solution since it was similar to our prior API and it follows the convention for plugin registration. Also, I could have implemented a caching technique with `container.getAll`, but there didn't seem to be noticeable overhead by calling `container.getAll` every time to find the correct filter  

- [ ] After this pull request is approved I will update the CustomFilter wiki from `registerSingleton` to `registerTransient`
- [ ] Add dependency injection with CustomFilter example


ghiscoding:
Looks good at first glance, I will review it more tonight after work. Though I have some quick notes
- I read somewhere that TypeScript now discouraged adding the `types` in javadoc, so `@param {string} inputStr` should just be `@param input string`
  - the reason is simple, we have intellisense with TypeScript so there's no need to add it again to the jsdoc
  - there is apparently a flag also in TSConfig that if enabled will fail on these `{type}` in curly braces
- IE and Edge are the ones that are more impacted on performance side, for example the Example 4, if I drag & drop a column position, it takes 2sec in Edge to re-render the grid, but in Chrome/Firefox it's instant. I'm really surprised that Edge is still impacted by performance in SlickGrid, I usually see these perf hit when dragging a column or grouping thousand of rows.

jmzagorski:
So is this performance impact you see because of the change? I thought once the grid is rendered that Factory does not need to be called again. Or are you saying "in gerernal" those two browsers perform slower?

ghiscoding:
I wouldn't go with caching, I think it's overkill and pron to caching issues. 

No the perf hit was always there in IE, I'm really surprised it's still there in Edge but anyhow it's still present. 

To give a bit more explanation on the Filters, I actually found a bug today in Angular/Aurelia, when you drag a column to another position, it will re-render the entire grid **and** the element filter. The bug I just fixed in Angular for now, is that the filter element (try on the multiple-select) loses his selected values. I'm not sure why it's slow in Edge, in theory SlickGrid only create DOM node for what is displayed on screen with a buffer of about 5 nodes on top/bottom, which is why when you scroll very far it's white at the beginning (that's the re-creating of DOM nodes that kicks in). I have a feeling that perf hit is because Edge/IE suck with arrays (like sort, reposition, filter, ...)

ghiscoding:
Reviewing your PR and the comment you posted in the issue
> What if we added a defaultFilterType to the SlickgridConfig that the user can override with their own configuration during the startup of the app when we pass the config object to them in main.ts here https://github.com/ghiscoding/aurelia-slickgrid/blob/master/aurelia-slickgrid/src/aurelia-slickgrid/index.ts#L66

I tested that part and I didn't know you could just use 
`import { SlickgridConfig } from '../slickgrid-config';`

That's cool, however, if I change my `main.ts` with
`config.options.defaultFilterType = FilterType.select;`

I was expecting that it might change all the input filters to a select but that didn't happen. Am I suppose to see something change or not? I did put a breakpoint to the Factory constructor and it does get the option of `FilterType.select`. It doesn't matter much for the fix, but I'm just curious to know what adding that option in the global config might do

jmzagorski:
ah good catch, I forgot to register your instance with the container. Add this after you `new` the `SlickgridOptions`:
```
  aurelia.container.registerInstance(SlickgridConfig, config);
```
if that works as you expected let me know and i can add that commit

ghiscoding:
Sorry, I'm not sure where you want me to add that line. Is that in the Factory or in the Config? 
BTW, it's not merged yet so any commit will follow

jmzagorski:
add it after https://github.com/ghiscoding/aurelia-slickgrid/blob/master/aurelia-slickgrid/src/aurelia-slickgrid/index.ts#L63. 

I'll be able to commit the change a little later tonight.

ghiscoding:
Ok I did that but still don't see any visual or behavior difference in the UI. The input filters remains input and they still work as an input.

ghiscoding:
We can do a separate PR if that helps. As long as we have the main fix resolved.

jmzagorski:
The changes in now should resolve the issue. To get that `defaultFilterType` to work I need to remove this line https://github.com/ghiscoding/aurelia-slickgrid/blob/master/aurelia-slickgrid/src/aurelia-slickgrid/services/filter.service.ts#L316

I did not see that before. Do you have any problem with that?

ghiscoding:
hmm ok, I guess you will pass the filterType in a different way to the switch/case?

jmzagorski:
Yeah sorry I didn't mean delete, i meant change it. Your filter service needs to know the filterType so that line would change to 

```
const filterType = (columnDef.filter && columnDef.filter.type) ? columnDef.filter.type : this._gridOptions.defaultFilterType;
```

which means your filter service will never pass an unknown or null filter type to `FilterFactory`, but the `FilterFactory` does not know that so it is a good safe guard in the `FilterFactory` to still have the `defaultFilterType`.

I did find one odd thing though. `SelectFilter` does not have a `filterType` property when I set a break point in the filter factory? I set the defaultFilterType. Do you see that too?

ghiscoding:
Right I just found the problem, my VSCode was also showing it in different color, the same color as a type... oh yeah you made it a type instead of an assignment :P  change the `:` to `=`

jmzagorski:
okay, new commit is in and that is it for me tonight. Hopefully these fix the problem. I'll catch up with everything tomorrow.

ghiscoding:
@jmzagorski 
Hey I just wanted to make sure, you're all done with the PR right? I ask bcoz you wrote 
> I'll catch up with everything tomorrow.

So just wanted to make sure before I merge. Thanks

jmzagorski:
yes, sorry. I meant I will catch up on the WIKI change and changes in my other future PRs

ghiscoding:
Ok thanks, I'll push a new version tomorrow or Monday. 
You can update the wiki on Monday, I'll put a note in the Project page

jmzagorski:
@ghiscoding 

Updated the wik page https://github.com/ghiscoding/aurelia-slickgrid/wiki/Custom-Filter. I will try to make my revision comments short and to the point https://github.com/ghiscoding/aurelia-slickgrid/wiki/Custom-Filter/_history

```diff
- Register your filter with the aurelia container 
+ Register your filter with the `registerTransient` method on the aurelia container 
```

```diff
- aurelia.container.registerSingleton(FILTER_PLUGIN_NAME, MyCustomFilter);
+ aurelia.container.registerTransient(FILTER_PLUGIN_NAME, MyCustomFilter);
```

Added a new section before **Default Search Term(s)**

### Default Filter Type
By default, the library uses the [inputFilter](https://github.com/ghiscoding/aurelia-slickgrid/blob/master/aurelia-slickgrid/src/aurelia-slickgrid/filters/inputFilter.ts) when none is specified. However, you can override this value with any filter you like during the startup/configuration of your Aurelia application:

**main.ts**
```typescript
  aurelia.use.feature(PLATFORM.moduleName('aurelia-slickgrid/index'), config => {
    // change any of the default global options
    config.options.defaultFilterType = 'myCustomFilter' // or you can use another library Filter
});
```

PR #29, opened 2018-03-11
fix(utilities): move findOrDefault to a function

### LINKED ISSUE #27: [bug]: findOrDefault shows in mergeGridOptions which uses jQuery extend

* **Bundler used (WebPack/RequireJS/SystemJS**
WebPack

* **Browser:**
all

* **Language:**
TypeScript

**Current behavior:**
Aurelia-Slickgrid `mergeOptions` function uses jQuery `$.extend()` and for some reasons the Grid Presets which before the merge has a valid array of objects, has the object replaced by a `findOrDefault` function.

No error and no Array prototype function

jmzagorski:
Having it on the Array.prototype causes issues with jquery extend and overall is risky. Also, and I am not sure how I missed this, but I was still using the `filter` property in the `collectionFormatter`. So i fixed that to make sure this new `findOrDefault` works.

closes #27

### DIFF
--- a/aurelia-slickgrid/src/aurelia-slickgrid/editors/singleSelectEditor.ts
+++ b/aurelia-slickgrid/src/aurelia-slickgrid/editors/singleSelectEditor.ts
@@ -4,7 +4,8 @@ import {
   Column,
   MultipleSelectOption,
   SelectOption
-} from './../models/index';
+} from '../models/index';
+import { findOrDefault } from '../services';
 import * as $ from 'jquery';
 
 /**
@@ -63,7 +64,7 @@ export class SingleSelectEditor implements Editor {
    * The current selected value from the collection
    */
   get currentValue() {
-    return this.collection.findOrDefault(c =>
+    return findOrDefault(this.collection, (c: any) =>
       c[this.valueName].toString() === this.$editorElm.val())[this.valueName];
   }
 
--- a/aurelia-slickgrid/src/aurelia-slickgrid/formatters/collectionFormatter.ts
+++ b/aurelia-slickgrid/src/aurelia-slickgrid/formatters/collectionFormatter.ts
@@ -1,26 +1,27 @@
 import { arrayToCsvFormatter } from './arrayToCsvFormatter';
 import { Column, Formatter } from './../models/index';
+import { findOrDefault } from '../services';
 
 /**
- * A formatter to show the label property value of a filter.collection
+ * A formatter to show the label property value of a params collection
  */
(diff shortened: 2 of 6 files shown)


ghiscoding:
I think you missed this [filter.collection](https://github.com/ghiscoding/aurelia-slickgrid/blob/master/aurelia-slickgrid/src/aurelia-slickgrid/formatters/index.ts#L52) as well, though it's just a comment. 

I was wondering what does the `collectionFormatter` does, so I went in the `index.ts` to see the description. Doesn't this look a lot like the `csvFormatter` that I made? At first glance, the description doesn't tell me the difference between the 2. Oh wait, actually nevermind, it takes the collection from the `params`, not from the value, that is the difference I guess.

jmzagorski:
the `collectionFormatter` is a lookup since the structure of the collection is `{ value: .., label... }`. For example if you had a collection like this `[{ value: 1, label: 'foo'}, {value: 2, label: 'bar' }]` and your dataset item on the grid had `[{ value: 1 },{ value: 2 }]` then the grid would display `foo` and `bar` and not `1` and `2`

ghiscoding:
Oh that is nice then, would you mind putting that example within the comment description of that formatter. I think it would be useful

jmzagorski:
do you want the description on the https://github.com/ghiscoding/aurelia-slickgrid/blob/master/aurelia-slickgrid/src/aurelia-slickgrid/formatters/index.ts#L53 or https://github.com/ghiscoding/aurelia-slickgrid/blob/master/aurelia-slickgrid/src/aurelia-slickgrid/formatters/collectionFormatter.ts

ghiscoding:
I've put all the description within the `index.ts` because those are the ones that show up with intellisense

jmzagorski:
I really need to get some form on intellisense or popup in my vim editor so I can see these jsdocs too. Not sure if what I just committed will look as you want it in your editor.

ghiscoding:
You might want to try [Vim for VSCode](https://github.com/VSCodeVim/Vim) extension, you could gain a lot from it without sacrificing your shortcuts and keys. Obviously the first thing you would gain is intellisense, which is 1 of the top 3 biggest reasons that I like TypeScript. 

Here's the intellisense for the previous code that was in place

code - insiders_2018-03-11_17-11-35

jmzagorski:
I used VS Code and VIM for VSCode a little over a year ago. I liked it a lot, but when I started getting more comfortable with VIM I found some quirks so I just rid of VSCode.  I will use VsVim in Visual Studio every now and then since I need it for work, but I find myself always going back to vim. VIM probably has something out there for typescript that I need to go find now that I am developing more in that language. Most of my plugins are javascript/c#.

PR #38, opened 2018-03-30
feat(filter/editor): add functionality to filter/sort collection

ghiscoding:
- we can now filter/sort the collection passed to the SingleSelect/MultipleSelect Editor and Filter

Tasks after merge & release
- [ ] update [Wiki](https://github.com/ghiscoding/aurelia-slickgrid/wiki/Select-Filter)


ghiscoding:
@jmzagorski 
I know you're busy, but If you have a chance to review, that would be great. Else I will merge and release on Monday/Tuesday.

ghiscoding:
yes that is correct, it's indeed an exclusion. In our App, we needed to sort the collection (also with translation), but I figured it wasn't that much harder to add filterBy as well... and I did :)

PR #43, opened 2018-04-08
feat(grid): add sample for Header Grouping spanning accros multiple cols

ghiscoding:


ghiscoding:
@jmzagorski 
You can review if you want and have time, I'll probably merge on Monday and look at a new version during the week before I leave in vacations. 

Some things worth noting about this PR, Grouping Header doesn't work so well with a window resize. Probably some CSS to fix up somewhere. Also, the pre-header and header-row have to be re-created every time the grid changes (like sorting, resizing column width, ...) which is dealt properly in FilterService with `grid.onHeaderRowCellRendered` which calls `addFilterTemplateToHeaderRow` but for the pre-header I found out there was a lot more to do on that side of it and it isn't yet correct with grid resize. 

I managed to do the following over the weekend
- add support for multi-columns number indicator & tristate (in master)
- add Column Span (colspan) demo (in master)
- add sample for Header Grouping spanning accros multiple cols (this PR)

I'd like to get this done too, which I'm about to start working on
- Copy & Paste like Excel, see [SlickGrid demo](http://6pac.github.io/SlickGrid/examples/example-excel-compatible-spreadsheet.html)

If you have something upcoming, let me know.

Issue #48, opened 2018-04-15
Conflict with Froala editor


haimavni:
**I'm submitting a bug report**

* **Framework Version:**
aurelia cli 0.27.0

**Please tell us about your environment:**
* **Operating System:**
Windows 7 and also Ubuntu 16.04

* **Bundler used RequireJS

* **Browser:**
Chrome
* **Language:**
 TypeScript

The problem is not directly with slickgrid. The problem is that when I add aurelia-slickgrid to the project dependencies, a plugin named aurelia-froala-editor is broken - it does not respond to mouse clicks at all.

ghiscoding:
Unfortunately I don't use that other npm package so you will have to troubleshoot yourself. 
what you can try is to manully delete each javascript file used by `Aurelia-Slickgrid` until you find that your other lib works. For example, you can start by deleting [jquery.event.drag-2.3.0.js](https://github.com/ghiscoding/aurelia-slickgrid/blob/master/aurelia-slickgrid/src/aurelia-slickgrid/aurelia-slickgrid.ts#L3) from the `slickgrid` folder. I'm  not sure how good this test will be though because it will probably stop build from the start. so that might be invalid. Anyhow, I don't know what else to try, I can't do much more

haimavni:
Hi Ghislain,

I am writing to you because I find slickedit very attractive and I wish I
could use it in my app.

I had another issue with aurelia-slickedit which is not related to the
other package. Also, the conflict with Froala editor may expose a more
general issue which is worth working on.
If you like to spend some time on these issues, please do the following:

git clone https://github.com/haimavni/test-grid.git
cd test-grid
npm install
au run --watch
In your browser, enter the address localhost:9000

It should not take more than a minute or two.

You will see the table showing nicely, as it should, and the editor is
there but you can not enter text.

If you uncomment the two commented lines in src/example1.html, you will see
that the screen remains empty and inspecting the console shows some error.
May be I am missing something but I don't think this should happen.

Please let me know if there are some (hopefully good) news.

Thanks,

Haim Avni

ghiscoding:
I'm in vacations for 3 weeks so no I won't test it. What I wrote if for you to test it. Please note that this lib is all done through free time and there isn't much more I can do at this point

haimavni:
Have a nice vacation!

בתאריך יום ב׳, 16 באפר׳ 2018, 14:46, מאת Ghislain B. ‏<
notifications@github.com>:

[...]

Issue #44, opened 2018-04-09
Sort icons on last column is hidden behind Grid Menu with small dataset


ghiscoding:
**I'm submitting a bug report**

* **Language:**
all

**Current behavior:**
Sort icons on last column is hidden behind Grid Menu with small dataset

**Expected/desired behavior:**

* **What is the expected behavior?**
Scrolling should probably be displayed at all time since the Grid Menu rely on the fact that the scrolling bar is the same width as the Grid Menu.

* **What is the motivation / use case for changing the behavior?**

ghiscoding:
See below for a demo of the issue.

I have opened an [issue #231](https://github.com/6pac/SlickGrid/issues/231) on SlickGrid and already created a [PR #232](https://github.com/6pac/SlickGrid/pull/232) to fix it. 

There's already a branch [bugfix/grid-menu-without-scrolling](https://github.com/ghiscoding/aurelia-slickgrid/commit/22514c683e15004617cbbbac3a472072c8cb3fa0). 

So it's only a matter of waiting to see if the PR gets approved in the next version of SlickGrid and coordinating the use of that newer release in our lib. 

![gridmenu_smalldataset](https://user-images.githubusercontent.com/643976/38507559-da7d7f68-3bea-11e8-8c82-96662b763481.png)

ghiscoding:
closed by PR #53

Issue #36, opened 2018-03-24
[feature] Multiple Grids in one view


jmzagorski:
**I'm submitting a feature request**

* **Framework Version:**
2.0 potentially or 1.x if not breaking

**Please tell us about your environment:**
* **Operating System:**
All

* **Browser:**
all

* **Language:**
TypeScript

**Current behavior:**
No supported since services are singletons and hold onto grid state

* **What is the expected behavior?**
Services can act on multiple grids

* **What is the motivation / use case for changing the behavior?**

Copying over the discussion from https://github.com/ghiscoding/aurelia-slickgrid/pull/33 to keep track of progress on this:

Creating a singleton stateful service has some drawbacks, and we discussed one of those yesterday with multiple grids on the same page. I can think of 3 options if you wanted to change your services, both of which would fix the multiple grid problem because it would not have one state.

1. Inject all grid services as transient. This means every slickgrid custom element will get an new instance of the service and work with just with that grid. The downside of this is that these services cannot be used in view models since a new instance is created every time it is injected. However, it would fix the multiple grids on a page issue

2. Use an object on all services to keep track of which grids are passed into init by either using the object itself or the grid id. This could get complicated because every method call to a service would have to pass some key so the service knows which grid to operaton on. Since you would make to pass in a key anyway as an additional argument for this to work, number 3 would be a better choice.

3. Make a stateless services. By no state I mean they do not have properties for the grid or anything the custom element created at runtime, they just perform a service on the grid, dataview, columns or whatever is passed into their methods. For example this line would pass the grid and row number as parameters. Users can still use your service in their view models because you expose grid objects anyway and if a method required the grid they could pass it. This would allow multiple grids to be rendered on the same page too.

After typing all of that, number 1 would be the least amount of change, but break use of your services outside of the custom element. I would not recommend number 2. Number 3 is another good option because services should be stateless, but would required a rewrite of all service methods requiring a grid object.

number 3 would have to be in version 2 and most likely they get the grid from some event or leave it exposed as a bindable.

I think number 1 is just as good and didn't want to rule it out. It potentially would not add any breaking changes while fixing the multiple grid problem (i think). We would need to think of a way to expose your service methods with 1 or many grids, which is challenging if we inject the services with @transient. Based on the index file that would be quite a few service methods.

jmzagorski:
[...]

PR #50, opened 2018-04-18
fix(editors): use indexOf in multiple select editor to load value

jmzagorski:
`indexOf` must be used because `defaultValue` is an array for the `multipleSelectEditor`. the current `multipleSelectEditor` is broken and values are not loaded when the multiple select list is opened.

This reverts part of this [commit](https://github.com/ghiscoding/aurelia-slickgrid/commit/71916dc8bd80e257c310fa89485fc8a2e534a5d1) for the `multipleSelectEditor`.

The excel export and the sort still works with numbers. However, the current state of sorters will not sort "Task 2" and "Task 10" correctly regardless of the editors because those values are strings and not numbers. A separate issue would have to be opened to that. I updated `Example3.ts` to show that problem by allowing in Tasks greater than 10 in the prerequisite column.

If I can figure out/have permission to release a hot fix (version 1.12.3) I will do that because I would imagine this is critical since it breaks inline editing.

### DIFF
--- a/aurelia-slickgrid/src/aurelia-slickgrid/editors/multipleSelectEditor.ts
+++ b/aurelia-slickgrid/src/aurelia-slickgrid/editors/multipleSelectEditor.ts
@@ -106,7 +106,7 @@ export class MultipleSelectEditor implements Editor {
     }
 
     // user might want to sort the collection
-    if (this.gridOptions && this.gridOptions.params && this.columnDef.params.collectionSortBy) {
+    if (this.columnDef.params && this.columnDef.params.collectionSortBy) {
       const sortBy = this.columnDef.params.collectionSortBy;
       newCollection = this.collectionService.sortCollection(newCollection, sortBy, this.enableTranslateLabel);
     }
--- a/aurelia-slickgrid/src/aurelia-slickgrid/editors/multipleSelectEditor.ts
+++ b/aurelia-slickgrid/src/aurelia-slickgrid/editors/multipleSelectEditor.ts
@@ -130,7 +130,7 @@ export class MultipleSelectEditor implements Editor {
     this.defaultValue = item[this.columnDef.field].map((i: any) => i.toString());
 
     this.$editorElm.find('option').each((i: number, $e: any) => {
-      if (this.defaultValue === $e.value) {
+      if (this.defaultValue.indexOf($e.value) !== -1) {
         $e.selected = true;
       } else {
         $e.selected = false;
--- a/aurelia-slickgrid/src/aurelia-slickgrid/editors/singleSelectEditor.ts
+++ b/aurelia-slickgrid/src/aurelia-slickgrid/editors/singleSelectEditor.ts
@@ -98,7 +98,7 @@ export class SingleSelectEditor implements Editor {
     }
 
     // user might want to sort the collection
-    if (this.gridOptions.params && this.columnDef.params.collectionSortBy) {
+    if (this.columnDef.params && this.columnDef.params.collectionSortBy) {
       const sortBy = this.columnDef.params.collectionSortBy;
       newCollection = this.collectionService.sortCollection(newCollection, sortBy, this.enableTranslateLabel);
     }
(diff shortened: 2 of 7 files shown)


ghiscoding:
Hmm it does revert my previous [commit](https://github.com/ghiscoding/aurelia-slickgrid/commit/71916dc8bd80e257c310fa89485fc8a2e534a5d1), have you tested the sort descending which was a fix in that previous commit? I'm a bit surprised since it's a complete revert of my previous commit, which was a fix. Anyhow, I don't use this inline editor in any of my projects yet, so you would more than I do on this subject.

For the string sorting, I would actually rather always sort by string as default. We might need another option to have another way to sort. If another option is created for that, we could take advantages of the built-in [Sorters](https://github.com/ghiscoding/aurelia-slickgrid/tree/master/aurelia-slickgrid/src/aurelia-slickgrid/sorters), these are used to sort data of local grid (without a Backend Service).

jmzagorski:
Would I be testing the data sorted on the grid or in the select list?

ghiscoding:
The previous commit I made was to fix the select list not being sorted correctly on a sort order descending. Does that answer your question?

jmzagorski:
Yes, thank you. I will double check the sorting DESC. Not sure how much work I will be able to do over the next week (i will be "off work", but may still be working lol), but this will be the next thing I will look into.

ghiscoding:
@jmzagorski 
Not totally related to this PR, but can you give an update on what has to be reviewed? 
I find there's a lot of open issues and I'd like to close a few, if possible. 

- For the current PR, I guess you will look into it on the short term
- issue #44 for the sort icon now has my PR/fix merged in SlickGrid, however waiting for a new release so that I can then apply the fix on this lib as well
- PR #51, I will take a look in the coming weekend to see if we can get rid of the function arguments all together, else will probably merge what you already submitted
- Issue #48 with Froala editor, that's a though one, not sure about it, that might be hard to fix since it might be a conflict with the jquery drag event lib
- issue #36, multiple grids in 1 view. This interest me a lot and I guess you will work on that one. The question is, will that be doable in version `1.x` or will that be in `2.x` ?

jmzagorski:
I reviewed example 3 again for sorting editors and added examples in 12 and 4 for filters. If you pull down the branch I can't find any problems with sorting.  Could I be missing something you saw? The `singleSelectEditor` was definitely wrong and you fixed that in the commit, but maybe the `multipleSelectEditor` was okay?

Also, the `multipleSelectFilter` and `singleSelectFilter` has a check for `this.gridOptions.params` [here](https://github.com/ghiscoding/aurelia-slickgrid/blob/master/aurelia-slickgrid/src/aurelia-slickgrid/filters/multipleSelectFilter.ts#L95) and I am not sure why since that is never used in the sort or filter.

Another thing to be aware of, if there is a DESC sortBy set on a filter, the blank default option will be thrown to the bottom. You can see this in my example 12 I added.

ghiscoding:
oh yeah I guess that is expected for a sort DESC lol. Not much we can do about that I think. Unless we fix it with some kind of patch but that is kind of ugly

jmzagorski:
I removed the `gridOptions` check, but kept the check on the `columnDef`

ghiscoding:
Oh yeah that does look good. Where you done with the PR then, anything else to review before the merge? I'm good with your changes but I didn't test locally and I don't think I need to since you tested yourself.

jmzagorski:
i am playing around with all the examples to see if an exceptions are thrown and if anything weird happens. After that I can merge it if you want.

Not related, do you notice the inline date editors not working? I select a date and it does not set it.

ghiscoding:
Yup go ahead and merge whenever ready 👍 

There's also this other [PR](https://github.com/ghiscoding/aurelia-slickgrid/pull/54) that you can merge after reviewing. That one, I was going to merge it myself but I know you use Inline Editor, so I left it open for you to be aware of, there's some description in it

jmzagorski:
Oh yea I saw that PR and added a review I never submitted last night. I submitted it now

Issue #35, opened 2018-03-22
[bug] - Grid height and width are not set


jmzagorski:
**I'm submitting a bug report**

* **Framework Version:**
1.10.0

**Please tell us about your environment:**
* **Operating System:**
Windows 10

* **Browser:**
Tested on Chrome/Firefox, but should be a bug on all

* **Language:**
Typescript

**Current behavior:**
The grid height and width are set using the `bindingContext` parameter on the `bind` method. However, the `bindingContext` will be my view model, which does not have a `gridHeight` or `gridWidth` properties because  I am setting the height and width directly in the view like so:

```html
<template>
  <aurelia-slickgrid 
    grid-id="elements-setup" 
    column-definitions.bind="elementColumns" 
    grid-options.bind="gridOptions" 
    dataset.bind="elements"
    grid-height="400" grid-width="800">
  </aurelia-slickgrid>
</template>
```

when inspecting `binding.gridHeight` and `binding.gridWidth`  [here](https://github.com/ghiscoding/aurelia-slickgrid/blob/master/aurelia-slickgrid/src/aurelia-slickgrid/aurelia-slickgrid.ts#L189), they are both undefined. They are undefined in Example1 too, so I am not sure how the height and width are being set to 800 and 400.

* **What is the expected behavior?**
Should use `this.gridHeight` and `this.gridWidth` since those will always have the values. Also you can probabaly remove `gridWidthString` and `gridHeightString` properties since they are not in use

jmzagorski:
Actually you are resizing the grid [here](https://github.com/ghiscoding/aurelia-slickgrid/blob/master/aurelia-slickgrid/src/aurelia-slickgrid/aurelia-slickgrid.ts#L386) so maybe I am confused why the `style` property is being used?

ghiscoding:
I can't fully recall, but doing the `${binding.gridHeight}px` might be because I tried to be in sync with my Angular repo. I haven't used much of the `binding` object from the `bind()` function yet. So yeah I could be totally off there. 

Your last comment with this [line](https://github.com/ghiscoding/aurelia-slickgrid/blob/master/aurelia-slickgrid/src/aurelia-slickgrid/aurelia-slickgrid.ts#L386) might have been added because the `${binding.gridHeight}px` wasn't working. 

[...]

PR #52, opened 2018-04-27
fix(gridOptions): use this.gridOptions in bind fn

jmzagorski:
I know this is small pull request, but I like to break it out so you can see the change without having to go looking in master commits....

Using the first arg in the bind method looks at the parent context, which might not be the view-model with the `gridOptions`. For example in a `repeat.for`, the view-model properties are not available, which can be wrong if aurelia slickgrid is in a `repeat.for`. Using `this.gridOptions` assures us we are using the options that are bound

### DIFF
--- a/aurelia-slickgrid/src/aurelia-slickgrid/aurelia-slickgrid.ts
+++ b/aurelia-slickgrid/src/aurelia-slickgrid/aurelia-slickgrid.ts
@@ -214,7 +214,7 @@ export class AureliaSlickgridCustomElement {
    */
   bind(binding: any, contexts: any) {
     // get the grid options (priority is Global Options first, then user option which could overwrite the Global options)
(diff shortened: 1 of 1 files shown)


ghiscoding:
That seems similar to the other issue we had in the past with the grid size.
So there won't be any use for the `binding` argument at all then.
 
Does this have any relations with the other issue you've opened #49 ?

jmzagorski:
It is similar to https://github.com/ghiscoding/aurelia-slickgrid/issues/35 because that issue used the bindingContext too. 

https://github.com/ghiscoding/aurelia-slickgrid/issues/49 is different and has to do with slickgrid caching the grid options after they are set.

ghiscoding:
We might as well remove entirely the arguments, 2 of them, since they aren't used at all. I'm ok with merging it after that change, thanks

ghiscoding:
Looks good, thanks

PR #55, opened 2018-05-12
refactor(services): remove gridOptions and columnDefinitions

ghiscoding:
- removed all references from all init/attach Services

**Note** it does not include the refactoring of `gridEventService.attachOnCellChange` and `gridEventService.attachOnClick` since that was already done in a separate PR #51 

@jmzagorski 
It's a long time that I wanted to cleanup these arguments and only use `grid` and `dataView` objects. We can get `options` and `columnDefinitions` from the `grid` and so we should do that everywhere.

I tested all the grids with all features and everything seems to work as usual.
Hopefully this will help in issue #36 (multiple grids in one view)

Note, this PR has to be merged **only after** PR #51 , hopefully without conflicts afterward but it's possible and will address at that time if that happens.

### DIFF
--- a/aurelia-slickgrid/src/aurelia-slickgrid/aurelia-slickgrid.ts
+++ b/aurelia-slickgrid/src/aurelia-slickgrid/aurelia-slickgrid.ts
@@ -132,7 +132,7 @@ export class AureliaSlickgridCustomElement {
     }
     this.controlAndPluginService.createPluginBeforeGridCreation(this.columnDefinitions, this.gridOptions);
     this.grid = new Slick.Grid(`#${this.gridId}`, this.dataview, this.columnDefinitions, this.gridOptions);
-    this.controlAndPluginService.attachDifferentControlOrPlugins(this.grid, this.columnDefinitions, this.gridOptions, this.dataview, this.groupItemMetadataProvider);
+    this.controlAndPluginService.attachDifferentControlOrPlugins(this.grid, this.dataview, this.groupItemMetadataProvider);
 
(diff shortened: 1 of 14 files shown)


ghiscoding:
In that case, would you mind pulling the branch and adding the `_gridOptions` in the `init` as what you think is ideal or provide the line and code to change. Thanks

jmzagorski:
I commit the changes to the services, see if it works for you. All i did was move the `_gridOptions` to a readonly property that always gets the grid options from slickgrid instead of a local field. 

The test i ran (not shown in the code) was to set a timeout in example11 that changes the `enableColumnPicker` to false. Then I clicked `Add New Mocked Item` with a debugger breakpoint in `highlightRow` to inspect the _gridOptions` and see it was false which means the options are up to date. Sort of a silly test since I know the readonly _gridOptions always calls the `this._grid.getOptions`, but I always try to test.

ghiscoding:
Oh wow ok, it's more than 1 line of code, I'm ok with the changes but I would like to test all examples before merging it. I'll do that in the coming days.

ghiscoding:
I left some comments because I'm at work and I'm trying to apply the changes to my Angular-Slickgrid as well. I also saw that the new getter
```javascript
private get _gridOptions(): GridOption {
    return (this._grid && this._grid.getOptions) ? this._grid.getOptions() : {};
  }
```
is not applied to all Services but it probably should be. So far I see these Services missing it
- [ ] `GridStateService`
- [ ] `GridOdataService`
- [ ] `GraphqlService`
- [ ] `ResizerService`

jmzagorski:
sorry just saw this. Looks good!

PR #58, opened 2018-05-16
fix(editors): fix serialization/deserilization in editors

### LINKED ISSUE #56: Date Editor is buggy again

**Current behavior:**
Doesn't seem to display days properly. Also choosing a date in the picker doesn't reflect the same date in the grid when it closes. I wonder if it's yet another `flatpickr` update issue, it always seem to have problem with TypeScript

The picker should work accordingly

jmzagorski:
`dateEditor` was serializing/deserializing wrong value. Also need to check for empty date value because moment will throw an error. `singleSelect` and `multipleSelectEditor` were serializing the actual value (e.g number type) when it should be serializing to a string and `applyValue` should be deserializing from string to actual type (e.g. number).

closes #56

### DIFF
--- a/aurelia-slickgrid/src/aurelia-slickgrid/editors/dateEditor.ts
+++ b/aurelia-slickgrid/src/aurelia-slickgrid/editors/dateEditor.ts
@@ -91,14 +91,22 @@ export class DateEditor implements Editor {
   }
 
   serializeValue() {
+    const domValue: string = this.$input.val();
+
+    if (!domValue) return '';
+
     const outputFormat = mapMomentDateFormatWithFieldType(this.args.column.type || FieldType.dateIso);
-    const value = moment(this.defaultDate).format(outputFormat);
+    const value = moment(domValue).format(outputFormat);
 
     return value;
   }
 
   applyValue(item: any, state: any) {
-    item[this.args.column.field] = state;
+    if (!state) return;
+
+    const outputFormat = mapMomentDateFormatWithFieldType(this.args.column.type || FieldType.dateIso);
+
+    item[this.args.column.field] = moment(state, outputFormat).toDate();
   }
 
   isValueChanged() {
--- a/aurelia-slickgrid/src/aurelia-slickgrid/editors/singleSelectEditor.ts
+++ b/aurelia-slickgrid/src/aurelia-slickgrid/editors/singleSelectEditor.ts
@@ -110,7 +110,8 @@ export class SingleSelectEditor implements Editor {
   }
 
   applyValue(item: any, state: any): void {
-    item[this.args.column.field] = state;
+    item[this.args.column.field] = findOrDefault(this.collection, (c: any) =>
+      c[this.valueName].toString() === state)[this.valueName];
   }
 
   destroy() {
@@ -119,7 +120,8 @@ export class SingleSelectEditor implements Editor {
 
   loadValue(item: any): void {
     // convert to string because that is how the DOM will return these values
-    this.defaultValue = item[this.columnDef.field].toString();
+    // make sure the prop exists first
+    this.defaultValue = item[this.columnDef.field] && item[this.columnDef.field].toString();
 
     this.$editorElm.find('option').each((i: number, $e: any) => {
       if (this.defaultValue === $e.value) {
@@ -133,7 +135,7 @@ export class SingleSelectEditor implements Editor {
   }
 
   serializeValue(): any {
-    return this.currentValue;
+    return this.$editorElm.val();
   }
 
   focus() {
--- a/aurelia-slickgrid/src/aurelia-slickgrid/editors/multipleSelectEditor.ts
+++ b/aurelia-slickgrid/src/aurelia-slickgrid/editors/multipleSelectEditor.ts
@@ -118,7 +118,9 @@ export class MultipleSelectEditor implements Editor {
   }
 
   applyValue(item: any, state: any): void {
-    item[this.args.column.field] = state;
+     item[this.args.column.field] = this.collection
+      .filter(c => state.indexOf(c[this.valueName].toString()) !== -1)
+      .map(c => c[this.valueName]);
   }
 
   destroy() {
@@ -141,7 +143,7 @@ export class MultipleSelectEditor implements Editor {
   }
 
   serializeValue(): any {
-    return this.currentValues;
+    return this.$editorElm.val();
   }
 
   focus() {


jmzagorski:
@ghiscoding i was re-reading the `serializedValue` method description [here](https://github.com/mleibman/SlickGrid/wiki/Writing-custom-cell-editors) again and I had once thought serializedValue should always be a string (because that is normally the result of serilization), but the description says it can be a simple arbitrary value. Do you think I should have left the serialized value the actual type from the original collection as opposed to returning the string value(s) from the `$input`? for the `singleSelectEditor` and `multipleSelectEditor`? For example, before if the `valueName` value on the `item` was 2 i serialized to 2 (same type), but with this PR i serialize to "2".

aurelia-slickgrid/src/aurelia-slickgrid/editors/dateEditor.ts
```diff
@@ -91,14 +91,22 @@ export class DateEditor implements Editor {
   }
 
   serializeValue() {
+    const domValue: string = this.$input.val();
```
ghiscoding: why not just `const domValue: string = this.$input.val() || '';` ?
That is what I've done in my other repo

aurelia-slickgrid/src/aurelia-slickgrid/editors/dateEditor.ts
```diff
@@ -91,14 +91,22 @@ export class DateEditor implements Editor {
...
+
     const outputFormat = mapMomentDateFormatWithFieldType(this.args.column.type || FieldType.dateIso);
-    const value = moment(this.defaultDate).format(outputFormat);
+    const value = moment(domValue).format(outputFormat);
 
     return value;
   }
 
   applyValue(item: any, state: any) {
-    item[this.args.column.field] = state;
```
ghiscoding: hmm is that necessary? I thought the `state` was already formatted by the serializer

aurelia-slickgrid/src/aurelia-slickgrid/editors/singleSelectEditor.ts
```diff
@@ -119,7 +119,8 @@ export class SingleSelectEditor implements Editor {
 
   loadValue(item: any): void {
     // convert to string because that is how the DOM will return these values
-    this.defaultValue = item[this.columnDef.field].toString();
+    // make sure the prop exists first
+    this.defaultValue = item[this.columnDef.field] && item[this.columnDef.field].toString();
```
ghiscoding: oh I learned something new, I didn't know we could write it this way. I always write the long way of doing the check of the property then do a ternary operator

aurelia-slickgrid/src/aurelia-slickgrid/editors/dateEditor.ts
jmzagorski: if `domValue` is anything other than a date (e.g. null, undefined or '') and you pass that to moment, the returned serialized value will be 'Invalid Date'

aurelia-slickgrid/src/aurelia-slickgrid/editors/dateEditor.ts
jmzagorski: i guess it depends how you want to handle this. Currently `serailizedValue` method returns a string date using `moment(date).format(string)`. If you used just the `state` value in `applyValue` then the original property for this editor will go from a date object type to a string type, and I do not think you want that? If you do not want to serialize it as a string, then this logic can move into the `serializedValue` method so the return value is a date object and then `applyValue` will go back to `item[this.args.column.field] = state;`

ghiscoding: ah right I didn't think about that, it is a string, I guess your change is fine then

ghiscoding:
I think your code looks good, you can merge whenever you are ready. I'm not sure if you were done with it or not.

jmzagorski:
I am. I just wanted to make sure you were okay with my responses to your comments and if they make sense?

ghiscoding:
yeah it looks ok

aurelia-slickgrid/src/aurelia-slickgrid/editors/dateEditor.ts
jmzagorski: so do you think it is right to serialize the date as a string and then deserilize in `applyValue`. The only downside I can see with that is that if someone overrides the `editCommandHandler`, the serializedValue will be a string and not a date object. Per the slickgrid documents he describes the function of `serializedValue` as:

> 
> // return the value(s) being edited by the user in a serialized form
>         // can be an arbitrary object
>         // the only restriction is that it must be a simple object that can be passed around even
>         // when the editor itself has been destroyed

And i think a date is a simple object that can be passed around after the editor is destroyed right?

jmzagorski:
I added one more comment on the `dateEditor` to get your opinion

aurelia-slickgrid/src/aurelia-slickgrid/editors/dateEditor.ts
ghiscoding: you can add a variable in the `onChange` to keep actual date and if never changed then just return the `defaultDate`. So something like this 
`return this.currentDate || this.defaultDate;`

ghiscoding: We need to make the Excel Copy Buffer Plugin working before we can say that the Editor works correctly

jmzagorski: I got it to work by adding `this.flatInstance.setDate(item[this.args.column.field])` in `loadValue` right after you set the `defaultDate`. However, I had a few questions before i commit my changes:

1. Is there a reason why you do not destroy flatpickr? This was commented out and instead you close it. However, it seems on init you are resetting the instance with a new element anyway.
2. You check for flatpickr with this code:

`this.flatInstance = (flatpickr && this.$input[0] && typeof this.$input[0].flatpickr === 'function') ? this.$input[0].flatpickr(pickerOptions) : null;`

Was there a reason you did not import flatpickr like `import flatpickr from 'flatpickr` and then initialize it like `this.flatInstance = flatpickr ? flatpickr(this.$input.get(0), pickerOptions) : null;`. In their documents it says this is the preferred way if their library is in a framework.

ghiscoding: The `flatpick` library is not that great with how to import it in TS, they changed it a few times too in trying to make it work in TS. Also it was the only way I got it working with both `WebPack` and RequireJS (the client-cli demo I have). So please don't change the import, it works now, don't break it again lol

I think I removed the delete of the `flatInstance` because SlickGrid destroys the DOM element even before `flatInstance`. Might not be exactly that, but it was something near that behavior.

ghiscoding:
@jmzagorski 
added a comment to your question on top

aurelia-slickgrid/src/aurelia-slickgrid/editors/dateEditor.ts
jmzagorski: thanks for that clarification, I keep forgetting about the CLI/RequireJS. I pushed the commit that gets the date to work with the excel cell copy

jmzagorski:
@ghiscoding off topic, but i didn't know where to say this. If you pull down the latest master and run your demo, are the inline editors working in example 3? All of them but the longtext and date are not working for me.

jmzagorski:
Oh yeah i meant to tell you this, if you `enableCheckboxSelector`, some inline editors stop working. I noticed this in my application, but I removed the checkboxselector and forgot about this problem.  In example 3 if you comment out `enableCheckboxSelect: true` in gridOptions, the editors work again

ghiscoding:
@jmzagorski 
Probably related to SlickGrid issue [#68](https://github.com/6pac/SlickGrid/issues/68)

I had a [PR #241](https://github.com/6pac/SlickGrid/pull/241) that got merged in last SlickGrid version `2.3.18` but I'm not sure if it will be reverted or not, it's causing some issue [#242](https://github.com/6pac/SlickGrid/issues/242) 

Anyhow, you need version `2.3.18` for it to work, if you don't then Row Selection and Editors aren't working properly together. The version upgrade is in `master` now, but you might not have it in this branch and/or you didn't upgrade your local

jmzagorski:
ahh i had a feeling i had something out of date. I will check that out later tonight or tomorrow

As for this PR, it should be ready.

ghiscoding:
Are you done with the PR? If so, please merge and close the issue. 

Also note, there is yet a newer version of SlickGrid `2.3.19` with a new flag to partially reverse a PR I made in previous version. I will update the version in a later commit

ghiscoding:
@jmzagorski 
Another side note to the (Row Selection + Inline Editors) used at same time. I found out that you cannot use the Excel Copy Buffer plugin at the same time. Because the row selection and excel copy both use the `Slick.CellSelectionModel` and one cancels the other. I added a note in the [Wiki - Row Selection](https://github.com/ghiscoding/aurelia-slickgrid/wiki/Row-Selection#description), let me know if that is not enough in the docs.

jmzagorski:
👍 looks good, thank you.

ghiscoding:
FYI, SlickGrid is now at version `2.3.19` and was updated on our side in this [commit](https://github.com/ghiscoding/aurelia-slickgrid/commit/1da7b7a1cb8f1ea1fdca3d88b8203a9525ccba21)

Issue #42, opened 2018-04-03
Change defaultBinding on dataset property in aurelia-slickgrid custom element


jmzagorski:
**I'm submitting a feature request**

* **Framework Version:**
current

**Please tell us about your environment:**
* **Operating System:**
All

* **Browser:**
all

* **Language:**
TypeScript X.X

**Current behavior:**
AureliaSlickgridCustomElement has a defaultBinding of `twoWay` on the `dataset` property in https://github.com/ghiscoding/aurelia-slickgrid/blob/master/aurelia-slickgrid/src/aurelia-slickgrid/aurelia-slickgrid.ts#L70

**Expected/desired behavior:**

* **What is the expected behavior?**
Do not specify a default binding, let it "default" to aurelia's default of `oneWay`

* **What is the motivation / use case for changing the behavior?**
I do not think we need the overhead of `bindingMode.twoWay` for the `dataset` property. The custom aurelia-slickgrid element will never update the dataset, but only get updates from the view model. Besides the slight overhead, there can be issues as I ran into. If a user is using the default `dataset.bind` in a view and that `dataset` is a getter property on the view model, the application will throw this error:

`ERROR [app-router] TypeError: Cannot set property newDataset of #<Example4> which has only a getter`

This issue is probabaly an edge case, but it does produce an Aurelia error that might be hard to debug for users. The work around is to make use `dataset.one-way` in the view, which will override the custom elements `defaultBinding`, but some users might not realize this error is coming from this library.

I can replicate this in example4.ts with the diffs below:
```diff
<template>
  <h2>${title}</h2>
  <div class="subtitle" innerhtml.bind="subTitle"></div>

  <aurelia-slickgrid grid-id="grid1" dataview.bind="dataview" grid.bind="gridObj" column-definitions.bind="columnDefinitions"
    grid-options.bind="gridOptions" 
- dataset.bind="dataset" 
+ dataset.bind="newDataset" 
asg-on-grid-state-service-changed.delegate="gridStateChanged($event.detail)">
  </aurelia-slickgrid>
</template>
```

```diff
- import {  autoinject } from 'aurelia-framework';
+ import { computedFrom, autoinject } from 'aurelia-framework';
import { CustomInputFilter } from './custom-inputFilter';
import { Column, FieldType, FilterType, Formatter, Formatters, GridOption, GridStateService } from '../../aurelia-slickgrid';

function randomBetween(min, max) {
[...]

Issue #57, opened 2018-05-15
Dynamically Add/Change Column Definitions and Grid Options


ghiscoding:
**I'm submitting a feature request**

**Current behavior:**
Not currently available

**Expected/desired behavior:**

* **What is the motivation / use case for changing the behavior?**
I want to be able to change/add Column Definitions dynamically and/or change Grid Options dynamically

ghiscoding:
This was brought up in my `Angular-Slickgrid` repo and is now supported. I am assuming we can also do this in this repo.

Some references
- [issue #20](https://github.com/ghiscoding/Angular-Slickgrid/issues/20) on the `Angular-Slickgrid` repo
  - a few notes and test cases are shown in there
- in that repo, there's a Getter on `columnDefinitions` which will trigger a `translateHeader()` if the Column Definitions changes, see [here](https://github.com/ghiscoding/Angular-Slickgrid/blob/master/src/app/modules/angular-slickgrid/components/angular-slickgrid.component.ts#L82) and [here](https://github.com/ghiscoding/Angular-Slickgrid/blob/master/src/app/modules/angular-slickgrid/components/angular-slickgrid.component.ts#L430)
   - we probably need to do the same in Aurelia

Reference PR #55 this PR is what will make this possible.

ghiscoding:
closed with PR #65

PR #65, opened 2018-05-19
feature(grid): ability to dynamically add or change Column Headers

ghiscoding:
- We can dynamically add or change column definitions, see Example12 for a demo
- reference to issue #57

### DIFF
--- a/aurelia-slickgrid/src/aurelia-slickgrid/aurelia-slickgrid.ts
+++ b/aurelia-slickgrid/src/aurelia-slickgrid/aurelia-slickgrid.ts
@@ -87,10 +87,10 @@ export class AureliaSlickgridCustomElement {
 
   @bindable({ defaultBindingMode: bindingMode.twoWay }) columnDefinitions: Column[] = [];
   @bindable({ defaultBindingMode: bindingMode.twoWay }) element: Element;
-  @bindable({ defaultBindingMode: bindingMode.twoWay }) dataset: any[];
   @bindable({ defaultBindingMode: bindingMode.twoWay }) gridPaginationOptions: GridOption;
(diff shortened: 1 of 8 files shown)


ghiscoding:
@jmzagorski 
Looked at the open issues and saw that the issue #42 to change default binding on `dataset` might also expand to current PR. If so, can you make the changes or tell me what to change. Thanks

jmzagorski:
with #42 all you have to do is change this:

```diff
-@bindable({ defaultBindingMode: bindingMode.twoWay }) dataset: any[];
+@bindable dataset: any[];
```

The reason: `twoWay` implies that `AureliaSlickGridCustomElement` will or can update the `dataset` and the changes will be propagated back to the view model (hence the name two-way). I believe, by default, custom elements have a binding of `oneWay` (from the view model to the custom element).

I do not think we will ever be need to mutate the dataset...right?

jmzagorski:
will now close #42

ghiscoding:
> I do not think we will ever be need to mutate the dataset...right?

but technically, a user could come and change the JSON dataset, I actually also do that with the OData backend service I think. We also need to support data insert, in that case we push to the dataset/dataView. Unless I misunderstood, don't we still need a two way binding for that to work?

jmzagorski:
It is fine if the user/viewmodel changes the dataset since those changes will always flow into the AureliaSlickgridCustomElement, but if AureliaSlickGridCustomElement class changes the dataset  (ie changes the instance to another variable) then that change won't update the viewmodel dataset. As you found out in the examples everything should still work as expected

Nothing more in the pipeline from me right now.

ghiscoding:
ok good, so you said you can close #42 since it was incorporated in this PR ?

jmzagorski:
Yes #42 can be closed after this is merged

Issue #68, opened 2018-05-22
DateFilters in filter-conditions leave invalid/null dates in filter


jmzagorski:
**I'm submitting a bug report**
Unless this is by design?

* **Framework Version:**
1.13.0

* **Browser:**
all

* **Language:**
TypeScript

**Current behavior:**
An example is in the dateIsoFilterCondition, which will always filter for invalid dates as coded [here](https://github.com/ghiscoding/aurelia-slickgrid/blob/master/aurelia-slickgrid/src/aurelia-slickgrid/filter-conditions/dateIsoFilterCondition.ts#L8). (if the date is invalid, the row will not be filtered). So when a user filters for a specific date they will always get a data item with a invalid value in the date

**Expected/desired behavior:**

* **What is the expected behavior?**
invalid dates are filtered out of the result set

* **What is the motivation / use case for changing the behavior?**
If there are a large number of `NULL` dates these will be included in the filter and the filter will seem broken to the user.

ghiscoding:
I think that returning true will make it stay at same level while will make it move in the sort order

jmzagorski:
I think the sorting acts weird too. Look at my screenshot in your example4. I added a null date for a bunch of records and the sort does not group the null dates together. The second screenshot is the filter not working too:

![image](https://user-images.githubusercontent.com/8404957/40378132-a5846810-5dc0-11e8-871b-efd2a41d2eb2.png)


![image](https://user-images.githubusercontent.com/8404957/40378173-c317f004-5dc0-11e8-8403-77a2b2d2d17d.png)

ghiscoding:
I had some similar issues at the beginning in that same Example 4 with Description column, which is why I added some `null` in the dataset. I think the sort works correctly on string, maybe try to see if the Duration (number) is working when null are in the play. Also if the string is working correctly, then maybe there's some logic in that filter condition to make it work that the date doesn't have

jmzagorski:
per the [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) it looks like the date sorters need to return -1 or 1 when the date values are null. 

The code changes below will need to be applied to all date sorters

**Soters**
```diff
- if (!moment(value1, FORMAT, true).isValid() || !moment(value2, FORMAT, true).isValid()) {
-    return 0;
-   }
+  let diff = 0;

+  if (!moment(value1, moment.ISO_8601).isValid()) {
[...]

PR #69, opened 2018-05-24
fix(filters): return false when invalid date

jmzagorski:
Returning true on an invalid date will keep it on the grid when it should be filtered out.

associates #68
will be closed when sorting is fixed in a separate PR


jmzagorski:
yes, probably a good idea. Nice catch.

PR #70, opened 2018-05-24
fix(sorters): return -1 or 1 for invalid dates

### LINKED ISSUE #68: DateFilters in filter-conditions leave invalid/null dates in filter

jmzagorski:
Returning zero every time value1 or value 2 is invalid will result in an incorrect sort because it would not sort one value higher than the other (e.g if value1 was 2018-05-24 and value2 was null, these would be sorted as equals). We want invalid dates to be grouped together in the sort and this accomplishes that.

We could add additional logic such that if both were invalid that we return 0, but this commit does the job just fine. Since the logic was the same for every sorter, with the exception of format and strict, I created a utility function that took all the parameters

closes #68
associates to PR #69

### DIFF
--- a/aurelia-slickgrid/src/aurelia-slickgrid/sorters/sorterUtilities.ts
+++ b/aurelia-slickgrid/src/aurelia-slickgrid/sorters/sorterUtilities.ts
@@ -32,9 +32,9 @@ export function sortByFieldType(value1: any, value2: any, fieldType: FieldType, 
 export function compareDates(sortDirection, value1, value2, format, strict?: boolean) {
   let diff = 0;
 
-  if (!moment(value1, format, strict).isValid()) {
+  if (value1 === null || value1 === '' || !moment(value1, format, strict).isValid()) {
     diff = -1;
-  } else if (!moment(value2, format, strict).isValid()) {
+  } else if (value2 === null || value2 === '' || !moment(value2, format, strict).isValid()) {
     diff = 1;
   } else {
     const date1 = moment(value1, format, strict);
--- a/aurelia-slickgrid/src/aurelia-slickgrid/sorters/sorterUtilities.ts
+++ b/aurelia-slickgrid/src/aurelia-slickgrid/sorters/sorterUtilities.ts
@@ -1,5 +1,6 @@
 import { FieldType } from './../models/fieldType.enum';
 import { Sorters } from './index';
+import * as moment from 'moment';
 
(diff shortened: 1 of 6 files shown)


ghiscoding:
Looks good, however I was thinking to revisit this and completely remove `moment` from the equation. Using `moment.js` with 2 steps (1st date validity, 2nd date compare) is terribly slow compare to regular javascript `Date`. I just didn't have the time to revisit it, but over 1000 items, you can visually see less than a second with `Date` versus 2-3 sec with `moment`. I just don't know what is the best algorithm to validate a date without `moment`

ghiscoding:
@jmzagorski 
However there is a way to improve performance with `null` dates... you don't need `moment` to know it's invalid, just check for `null` in the same `if`

aurelia-slickgrid/src/aurelia-slickgrid/sorters/sorterUtilities.ts
```diff
@@ -27,3 +28,19 @@ export function sortByFieldType(value1: any, value2: any, fieldType: FieldType,
 
   return sortResult;
 }
+
+export function compareDates(sortDirection, value1, value2, format, strict?: boolean) {
+  let diff = 0;
+
+  if (!moment(value1, format, strict).isValid()) {
```
ghiscoding: add the `null` check in both `if` condition, this will improve a lot performance. Please try with 5000+ rows and you should see a drastic improvement with the `null` check

ghiscoding:
Added a `null` and empty string check, tested the Start Date column with 5000 rows. I see a 10x improvement with the `null` check (when there are null dates that is)
- with `null` check it took 1.2 sec to sort
- without `null` check it took 10.2 sec to sort

jmzagorski:
wow great testing! Thank you for finishing this up, i got busy yesterday with preparing a release for my application. Your way is smarter since the main reason for this PR was missing dates, Great work!

PR #76, opened 2018-06-06
fix(sort): Presets & Sort caused unintended behavior w/dataview changed

ghiscoding:
- fixes Presets unintended behavior after Clearing Sorting, then Clearing Filters. 
- also refactored dispatch of Custom Event to be 1 liner for cleaner code

@jmzagorski 
Thanks a lot in finding this one, I was pulling my hair.
The DataView subscribe was way overkill to do the presets, just calling the `loadLocalPresets` after the DataView finished updating is more than enough. That is much cleaner and it works all the time (one time that is... lol)

### DIFF
--- a/aurelia-slickgrid/src/aurelia-slickgrid/aurelia-slickgrid.ts
+++ b/aurelia-slickgrid/src/aurelia-slickgrid/aurelia-slickgrid.ts
@@ -125,9 +125,7 @@ export class AureliaSlickgridCustomElement {
   }
 
   initialization() {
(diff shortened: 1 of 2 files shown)


ghiscoding:
Yeah I was thinking why I used it and then I remembered that it was working when I create the Plugins simply because the DataView must exist and must be filled before you can use `setSortColumns`

Are you ok with the `dispatchCustomEvent` refactoring that is included in this PR?

jmzagorski:
yup that looks good

PR #79, opened 2018-06-09
feat(filter): add new slider filter (input range)

ghiscoding:
Use an input range to add what looks like a slider filter. 
- when the slider is set to 0 (default value), it returns all values
- when going over over, it will return only the rows with selected value
- styled to be cross-browser compatible
- tooltip only shown in IE, I wish we could add them for all browser though

@jmzagorski 
It's not yet finished but it's functional, so don't approve yet. I just need to clear up the styling and see if we can add tooltip for all browser. Not as useful without tooltip of the value

chrome_2018-06-09_18-35-56

### DIFF
--- a/aurelia-slickgrid/src/aurelia-slickgrid/filters/sliderFilter.ts
+++ b/aurelia-slickgrid/src/aurelia-slickgrid/filters/sliderFilter.ts
@@ -109,16 +109,24 @@ export class SliderFilter implements Filter {
     const maxValue = this.filterParams.sliderMaxValue || DEFAULT_MAX_VALUE;
     const step = this.filterParams.sliderStep || DEFAULT_STEP;
 
+    if (this.filterParams.hideSliderValue) {
+      return `
+        <input type="range" id="rangeInput_${this.columnDef.field}"
+          name="rangeInput_${this.columnDef.field}"
+          defaultValue="${defaultValue}" min="${minValue}" max="${maxValue}" step="${step}"
+          class="form-control search-filter range"
+          onmousemove="$('#rangeOuput_${this.columnDef.field}').html(rangeInput_${this.columnDef.field}.value)" />`;
+    }
+
     return `
-    <div class="input-group">
-      <input type="range" id="rangeInput_${this.columnDef.field}"
-        name="rangeInput_${this.columnDef.field}"
-        defaultValue="${defaultValue}" min="${minValue}" max="${maxValue}" step="${step}"
-        class="form-control search-filter range"
-        oninput="$('#rangeOuput_${this.columnDef.field}').html(rangeInput_${this.columnDef.field}.value)" />
-      <span class="input-group-addon" id="rangeOuput_${this.columnDef.field}">${defaultValue}</span>
-    </div>
-    `;
+      <div class="input-group">
+        <input type="range" id="rangeInput_${this.columnDef.field}"
+          name="rangeInput_${this.columnDef.field}"
+          defaultValue="${defaultValue}" min="${minValue}" max="${maxValue}" step="${step}"
+          class="form-control search-filter range"
+          onmousemove="$('#rangeOuput_${this.columnDef.field}').html(rangeInput_${this.columnDef.field}.value)" />
+        <span class="input-group-addon" id="rangeOuput_${this.columnDef.field}">${defaultValue}</span>
+      </div>`;
   }
 
   /**
--- a/aurelia-slickgrid/src/aurelia-slickgrid/filters/sliderFilter.ts
+++ b/aurelia-slickgrid/src/aurelia-slickgrid/filters/sliderFilter.ts
@@ -12,16 +12,16 @@ import * as $ from 'jquery';
 
 const DEFAULT_VALUE = '0';
 
-export class InputRangeFilter implements Filter {
+export class SliderFilter implements Filter {
   private $filterElm: any;
   grid: any;
   searchTerms: SearchTerm[];
   columnDef: Column;
   callback: FilterCallback;
 
-  /** Getter for the Grid Options pulled through the Grid Object */
-  private get gridOptions(): GridOption {
-    return (this.grid && this.grid.getOptions) ? this.grid.getOptions() : {};
+  /** Getter for the Filter Generic Params */
+  private get filterParams(): any {
+    return this.columnDef && this.columnDef.filter && this.columnDef.filter.params || {};
   }
 
   get operator(): OperatorType | OperatorString {
@@ -68,7 +68,7 @@ export class InputRangeFilter implements Filter {
    */
   clear() {
     if (this.$filterElm) {
-      const clearedValue = this.columnDef && this.columnDef.params && this.columnDef.params.sliderDefaultValue || DEFAULT_VALUE;
+      const clearedValue = this.filterParams.sliderDefaultValue || DEFAULT_VALUE;
       this.$filterElm.children('input').val(clearedValue);
       this.$filterElm.children('span.input-group-addon').html(clearedValue);
       this.$filterElm.trigger('change');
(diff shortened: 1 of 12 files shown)


ghiscoding:
oh oh I think I got what I wanted with Bootstrap input group, this is awesome. I use jQuery to update the appended value on the right. It seems to work on almost all browser, except IE 11 (as always) where the value doesn't get updated, however IE shows value as a tooltip so I'm fine with that. 

This is lovely 🏖

Still have to cleanup the CSS and make it better SASS and I'm done with it.

ghiscoding:
@jmzagorski 
The PR is ready for a review, however in the last commit I added multiple options in the generic `params`. However I'm really considering adding all of these options inside the `ColumnFilter` interface. I think these options can be used in a couple of other filters. Also, I see that we could also create 2 more filters (`numberFilter` and `floatFilter`) since we have these 2 as well inside the Editors. Does that make sense to do that or just these options inside the generic `params`? 

The options I was talking about are the following
- minValue
- maxValue
- step
- defaultValue
The latter might not be used outside of the slider (input range) though

aurelia-slickgrid/src/aurelia-slickgrid/filters/sliderFilter.ts
```diff
@@ -26,8 +26,13 @@ export class SliderFilter implements Filter {
...
   }
 
+  /** Getter for the `filter` properties */
+  private get filterProperties(): ColumnFilter {
+    return this.columnDef && this.columnDef.filter || {};
+  }
+
   get operator(): OperatorType | OperatorString {
-    return OperatorType.equal;
+    return this.columnDef.filter.operator || OperatorType.equal;
```
ghiscoding: @jmzagorski 
I had no choice but to code the `operator` this way. I tried to use the `this._operator` with the Getter/Setter like I saw in other filters (like the multipleSelect) but that ain't working. I might have misunderstood how it's use, but I don't see how the Setter could get called!? 

Passing the `operator` from the Column Definitions could be useful at some times (especially so for the Slider Filter, since we might want to use EQ for some grid but in other we might want '>=')

ghiscoding:
Also added a new Compound Slider Filter. To see all Slider features and options, you can see Example 9, 12 & 15 which uses different options in each. The Compound Filter is in Example 9

aurelia-slickgrid/src/aurelia-slickgrid/filters/sliderFilter.ts
jmzagorski: I looked at your current file and you have the right idea. If the operator can be set by the user (e.g. it is can `EQ` or `>=` then you need to define a getter and a setter similar to the compound filters. Other filters such as the multiple select only have a getter because the filter operator should only be `in` since any other operator does not makes sense. So if the user tries to set the operator on the `MultipleSelect` it should throw a readonly error. Hopefully that helps clear that up.

jmzagorski: Okay i see what you mean, I was looking at the `CompoundSliderFilter`. 

For the basic `SliderFilter`, the operator should only be equal (i think) because that is the default operator and it cannot be changed by the user at run time since it is not the `CompoundSliderFilter`. So you should not need a setter, only a getter:

```typescript
  get operator(): OperatorType | OperatorString {
    return OperatorType.equal;
  }
```

Unless I misunderstood some of these filters and the programmer can change the operator on non-compound filters, and if that is the case then some of the other filters might have to be changed (e.g. InputFilter will always be `EQ`)

ghiscoding: @jmzagorski 
Yes I do want to provide this flexibility of passing an operator, even non-compound filters. With that in consideration, is my code correct for acquiring the `operator` or should I change something? I still don't understand what and who calls the Setter though

jmzagorski: i think you want it to make the `SliderCompound` setter similar to the `CompoundSliderFilter`. 

```typescript
  set operator(op: OperatorType | OperatorString) {
    this._operator = op;
  }

  get operator(): OperatorType | OperatorString {
    return this._operator || OperatorType.empty;
  }
```

the operator on the Filter interface is really just a fallback because the ColumnFilter interface operator is used first. I did this because I needed a place to put the default operator since we removed filterType from the filter service. The filter service was using a switch statement to get the default operator based on the filter type, but I had to remove this logic. Putting on the instance was practical since a class should know what it default operator should be.

ghiscoding: but that's the thing, I did try to use the Getter/Setter that way and it didn't work. The `this._operator` is always `undefined` even though I set it inside the grid definition as `filter: Filters.Slider, operator: '>' }` 

You might need to build the project to see the behavior :P

jmzagorski: i see what you are saying, The filter service never actually sets the operator. So that means you had it correctly implemented and the other filters that have a getter only should also have that same logic i guess

Or you can create a getter only (similar to the InputFilter) with whatever default operator you want because I think your filter service will use the `columnDef.filter.operator` before checking the `columnDef.filter.model.operator`

ghiscoding: it does work the way I coded it, it pulls the `operator` if I passed it in the grid options and if I didn't then it uses the default one. I was just wondering if that is the best code. Considering what you wrote, I guess it's the best approach, is that correct?

How do you like the slider by the way? I always thought I needed an external lib to do that but found I could do it with just CSS styling :)

jmzagorski: yeah I like your logic in the getter, probabaly a good idea to change the other filters that do not have that logic too.

I really do like the filter. I may try it out in my app to get user feedback

ghiscoding: Ok then I'm done with this PR then. I want to create 2 other Filters anyway, I realized that the input filter should be split into `number` and `float` and perhaps make the `input be an alias to `text`. I can implement the getter logic on that other PR

Also, we should take this code and make an Editor as well

aurelia-slickgrid/src/aurelia-slickgrid/index.ts
```diff
@@ -44,6 +44,7 @@ export function configure(aurelia: any, callback: any) {
   aurelia.container.registerTransient(Filters.compoundDate);
   aurelia.container.registerTransient(Filters.compoundInput);
   aurelia.container.registerTransient(Filters.input);
+  aurelia.container.registerTransient(Filters.slider);
```
jmzagorski: do you need to register the compoundSlider too?

ghiscoding: good catch, I created the Compound Slider just last night and completely forgot to add it in there, I'll add it

ghiscoding: done

ghiscoding:
Hmm that is a valid point, the differences that exists between the 2 are basically that there's an extra DOM element for the operator to take care of (which has also triggers a filter change). There might be a way to reuse and merge some code, however I still want to keep the Filters as 2 distinct Filters for the end users.

So I think that the best we could do is create a common file to share code (like `sliderCommon.ts` and for all other Filters like Date & Input). Basically the Slider could end up being 3 files and the 2 filters would be smaller in size since the common would share the code. 

We can certainly do that in the future, but I'd do it in future PRs, it's low priority and doesn't affect at all the functionality. It only simplifies the code and it's size.

PR #87, opened 2018-07-28
feat(filter): add Prefix/Suffix/RenderHtml to Filters & Editors

ghiscoding:
Some new neat features for the MultipleSelect/SingleSelect Filters & Editors. 
- add `labelPrefix`, `labelSuffix` properties which allow using other properties of the `collection` as Prefix/Suffix
- add new `enableHtmlRender` to provide a way to render HTML
   - use of a new 3rd party lib to sanitize any bad attributes (like `<script>`) and avoid XSS

TODO
- [ ] update [Wiki](https://github.com/ghiscoding/aurelia-slickgrid/wiki/Select-Filter) with examples


ghiscoding:
@jmzagorski 
No worries, I put you as a reviewer if you have a chance to review though I know you are busy. So I decided to merge it after a few commits and after using it on our current project (always Angular that is). 

I already updated the Wikis for [Filter](https://github.com/ghiscoding/aurelia-slickgrid/wiki/Select-Filter#collection-label-prefixsuffix) and [Editor](https://github.com/ghiscoding/aurelia-slickgrid/wiki/Editors#collection-label-prefixsuffix), in fact this is all merged and released as of last weekend ;)

What it does is allowing you to combine multiple properties to show them as a combine text in each dropdown option labels. 

This PR also included a new property to [render HTML](https://github.com/ghiscoding/aurelia-slickgrid/wiki/Select-Filter#collection-label-render-html) in the same Filters/Editors. So now I can put in some Font-Awesome icons in my collection and they get displayed (provided that I enable the flag `

For example, say you have the following collection
```javascript
[ { symbol: '$', currency: 'USD', country: 'USA' }, { symbol: '$', currency: 'CAD', country: 'Canada' }];
```

You can display all of these properties inside your dropdown labels, say you want to show (symbol with abbreviation and country name). Now you can.

So you can create the  `multipleSelect` Filter and Editor with a `customStructure` by using the symbol as prefix, and country as suffix. That would make up something like this:
- $ USD USA
- $ CAD Canada

with a `customStructure` defined as 
```javascript
filter: {
      collectionAsync: this.http.get<{ option: string; value: string; }[]>(URL_SAMPLE_COLLECTION_DATA),
      collectionFilterBy: {
        property: 'value',
        operator: OperatorType.notEqual,
        value: 365
      },
      collectionSortBy: {
        property: 'value',
        sortDesc: true,
        fieldType: FieldType.number
      },
      customStructure: {
        value: 'currency',
        label: 'currency',
        labelPrefix: 'symbol',
        labelSuffix: 'country',
        addSpaceBetweenLabels: true
      },
          model: Filters.multipleSelect
}
```
This example is actually good, I should probably update the Wikis with that example scenario, it's understandable for everyone since we all need money lol

jmzagorski:
ahh that makes sense. Thank you for a great explanation. I have't needed to use HTML in these select lists (yet), and my workround to get a custom label with multiple properties was to use the `@computedFrom` decorator to combine the property values and then use that array for the select list. Glad to know I have other options now.

PR #89, opened 2018-08-13
feat(async): Filter & Editor load Collection Async & Watch, closes #83

### LINKED ISSUE #83: feature - support Editors & Filters with Async Collections

**Current behavior:**
Since this is currently not available, user need to resolve collections **before** creating the grid

Simply pass an `http` Promise to an async collection property and let it do it's job

Make it simpler to load asynchronously a collection and provide it to filter/editor without delaying the grid creation. 

#### TODO
- [x] Proof of Concept
- [x] Add a JSON file to simulate real `HttpClient` call for `collectionAsync`
- [x] Cleanup the code in MultipleSelectFilter
- [x] Extract duplicate functions/code in `MultipleSelectFilter`, `SingleSelectFilter` & `SelectFilter` and create a shared utility file or class
- [x] implement the same changes to the Editors `MultipleSelectEditor`, `SingleSelectEditor`
- [ ] update [Editor Wiki](https://github.com/ghiscoding/aurelia-slickgrid/wiki/Editors) & [Filter Wiki](https://github.com/ghiscoding/aurelia-slickgrid/wiki/Select-Filter)

ghiscoding:
This PR closes issue #83


ghiscoding:
@jmzagorski 
This would be great if you have a chance to quickly review, though I know it's a big PR and you're quite busy. I plan to release this next weekend.

Some highlights
- [Filter] all the code to deal with `collectionAsync` and collection watch are in the new `selectFilter.ts`
- [Editor] this is different, all the code for `collectionAsync` is in `aurelia-slickgrid.ts` since that is where we deal with the `internalColumnEditor` and so that is where I can control the collection load.
   - also note, editor does not need any collection observer, simply because even if we push to the collection, the next time we ready to create the editor element, it will have the new collection at the time.
   - however we do need to get the column reference pointer with `getColumns()` and reassign the `editor` to the `internalColumnEditor`
- to simplify the code (DRY) of Multiple/Single Editor/Filter, I merged all code into a file for each (`selectEditor`, `selectFilter`) which creates a much smaller file size
- because of previous point, I suggest you wait before creating any PR for the editor auto-save 
- finally the Example 3 (Editor) has all the code sample for both Editor & Filter (`collectionAsync` and Add/Delete item from collection). So I removed any sample from Example 4 (Filter), though I left `collectionAsync` in that one.

aurelia-slickgrid/src/aurelia-slickgrid/aurelia-slickgrid.ts
```diff
@@ -622,11 +626,51 @@ export class AureliaSlickgridCustomElement {
...
+
+  /** Load the Editor Collection asynchronously and replace the "collection" property when Promise resolves */
+  private loadEditorCollectionAsync(column: Column): any[] {
+    const collectionAsync = column && column.editor && column.editor.collectionAsync;
+    if (collectionAsync) {
+      // wait for the "collectionAsync", once resolved we will save it into the "collection"
+      // the collectionAsync can be of 3 types HttpClient, HttpFetch or a Promise
+      collectionAsync.then((response: HttpResponseMessage | Response | any[]) => {
+        if (response instanceof Response && typeof response['json'] === 'function') {
+          response['json']().then(data => this.updateEditorCollection(column, data));
```
jmzagorski: linters might warn or error here and recommed you use `response.json` over `response['json']. I know if you used the aurelia rules it would error

jmzagorski: inside the if 

```
if (response.bodyUsed) { 
  // do nothing or throw
} else response['json']().then(data => this.updateEditorCollection(column, data))

ghiscoding: what is the `bodyUsed` meant to do? As for the error, I think I would rather do nothing instead of throwing an error, can that ever happen anyway?

jmzagorski: `bodyUsed` checks to see if the response body was read from already. You can try it out in your code by calling `response.json` twice or i think if you call `response.text()` then `response.json()` too, both should error on the second call if i am not mistaken, (this would mimic what would happen if a user called `response.json` before your code),

jmzagorski: it can happen, but the user should protect against it if they know your code is reading it as well. Normally the user can prevent it by calling `response.clone()` first. I just did not know if you wanted to prevent a possible error from being thrown in your library.

ghiscoding: are you saying that we actually cannot call `.json()` twice? If so I didn't know and if that is the case then yes we should throw an error

ghiscoding: I would go with what you prefer since I'm not totally familiar with was is best to do here. If you could just provide me the lines of code then I'll update it tonight

jmzagorski: I can look more at it tomorrow and pull down ur code to double check what i have seen in the past. I would look at it tonight if I could, but I have not had much free time to code after work hours unfortunately

ghiscoding: just pushed another commit to change other things. BTW, since you will pull the code, I would say just push commit directly on the branch with the changes that make sense to you.

jmzagorski: just pushed the change with a comment for why in the commit message, sorry for the delay.

ghiscoding:
@jmzagorski 
I see that you made the change which is good, are done with the review? Are we ready to merge?

jmzagorski:
I am done. You can merge

ghiscoding:
@jmzagorski 
Thanks for the review, I'm planning a release in the weekend. 
Should I wait for the other issue of auto-save?

jmzagorski:
I'll have a better idea tomorrow. If I can finish a few backlog items at work, I can start the auto-save

ghiscoding:
@jmzagorski 
Awesome, but really there's no stress on my side, I can push with/without that other feature. I do need that feature in our current project for the futre, but this project is not due until end of September so I can wait. I can push new version of Aurelia-Slickgrid at any point in time, I'm quite flexible. 

Thanks for all

PR #94, opened 2018-09-08
fix(editor): Select Editor should close when tabbing away, fixes #92

### LINKED ISSUE #92: [Bug] Pressing tab while in single/multiple select editor leaves the select list open

* **Library Version:**
Current

* **Operating System:**
Windows 10

* **Browser:**
all
* **Language:**
TypeScript

**Current behavior:**
If the select list is open and a user hits tab, the select list remains open. If you try to select a value nothing happens since tab has activated the next cell

The select list closes

It is really not a big bug, but makes the user experience better if the select list went away or if tab did nothing, whichever you thought was better

ghiscoding:
- Single/Multiple Editors now closes the drop when tabbing away (blur)
- also fixed the `focus()` function inside the Editor
- demo shown below, I used tab key and arrows to test it out

- Since this seems to work quite good, I should probably do the same on the Select Filter as well. I'll add another commit within same PR

### DIFF
--- a/aurelia-slickgrid/src/aurelia-slickgrid/editors/longTextEditor.ts
+++ b/aurelia-slickgrid/src/aurelia-slickgrid/editors/longTextEditor.ts
@@ -54,11 +54,11 @@ export class LongTextEditor implements Editor {
         <button class="btn btn-default btn-xs">${cancelText}</button>
       </div>`).appendTo(this.$wrapper);
 
-    this.$wrapper.find('button:first').on('click', (event: Event) => this.save());
-    this.$wrapper.find('button:last').on('click', (event: Event) => this.cancel());
-    this.$input.on('keydown', this.handleKeyDown);
+    this.$wrapper.find('button:first').on('click', () => this.save());
+    this.$wrapper.find('button:last').on('click', () => this.cancel());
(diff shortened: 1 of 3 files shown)


jmzagorski:
thanks for doing this. I have not done much development on my app in some time since we are keeping it stable in production first. However, I do plan to migrate an old jquery/slickgrid app to this library so hopefully my priority can switch back to this lib as most of my development is during work hours now.

ghiscoding:
@jmzagorski 
Sure, does the PR look good to you then? You can merge after review.

Also, when do you think you might have a chance to fix the other editor auto-commit? Not that it's really rush, we are not shipping until mid-October I believe. So I'm just asking, or you can point me to the code

ghiscoding:
Ah I did not try that Editor, thanks for pointing it out. I'll review that tonight and possibly append to this PR with a commit if need be. Thanks for the review

ghiscoding:
Fix the error thrown in the console 

> longTextEditor.ts:76 Uncaught TypeError: Cannot read property 'grid' of undefined

The event handler was missing a `this` binding for it to work correctly. 
`this.$input.on('keydown', this.handleKeyDown.bind(this));`

PR #96, opened 2018-09-16
feat(editor): Add auto commit edit flag to grid options

### LINKED ISSUE #82: [Feature]: Option to auto commit inline editor when editor looses focus

* **Framework Version:**
current

* **Browser:**
all

* **Language:**
TypeScript

**Current behavior:**
Inline editors do not commit the value to the grid unless the user selects another cell or hits the enter/tab key. Some references to this issue include: 

https://coderwall.com/p/vztjkg/make-sure-slickgrid-commit-the-changes-when-click-on-submit-button-without-move-focus-out-of-the-input

https://github.com/mleibman/SlickGrid/issues/873

We could add a grid or editor option flag to commit on blur/focusout so the value in the editor is commited when the editor looses focus

It is more natural for a user to change the value in an inline editor and then navigate away from the grid to either save the values or do something else. I wanted your thoughts on this to see if it will be a useful feature for this library. I have not solved it in my app yet because the select editors act different when the 
`focusoutevent` ocurrs, which is the event i was listening for.

jmzagorski:
The auto commit flag will attempt to commit the current editor when the
editor is focused out. However, not all editors behave the same so extra
logic had to be included in some editors. This means that the grid may not
be able to auto commit custom editors and it will be the
responsibility of the user building the editor.

**checkboxEditor**: Can be handled by the focusout event, but a click event
was added in the editor to make it behave more like a checkbox
**dateEditor**, **sliderEditor**: Prevent focus to the next editor
**longTextEditor**: Add focusout event to textarea, grid does not capture event
**selectEditor**: No focusout event so commit on close
**others**: handled by aurelia-slickgrid subscription to focusout

Also, i tried to build the code and it is broken right now from what looks like other changes.

closes #82

### DIFF
--- a/aurelia-slickgrid/src/aurelia-slickgrid/aurelia-slickgrid.html
+++ b/aurelia-slickgrid/src/aurelia-slickgrid/aurelia-slickgrid.html
@@ -1,6 +1,7 @@
 <template>
   <div id="slickGridContainer-${gridId}" class="gridPane" css="width: ${gridWidth}px">
-    <div id.bind="gridId" class="slickgrid-container" style="width: 100%" css="height: ${gridHeight}px">
+    <div id.bind="gridId" class="slickgrid-container" style="width: 100%" css="height: ${gridHeight}px"
+      focusout.delegate="commitEdit($event.target)">
     </div>
 
     <slick-pagination id="slickPagingContainer-${gridId}" if.bind="showPagination" asg-on-pagination-changed.delegate="paginationChanged($event.detail)"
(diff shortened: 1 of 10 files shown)


ghiscoding:
Thanks it looks fine to me, however I didn't have a chance to run it locally yet. 

As for the Build, I mostly fix the warnings once I'm nearly ready for a release. Anyhow, I ran the Build now and fixed the warnings, you can run yours now.

jmzagorski:
still fails (image below). Want me to merge anyway?

image

ghiscoding:
@jmzagorski 
I made another commit on Master to address that, can you try merging again? 

Also, I'm sorry about the indentation, it seems to have been caused by VSCode (insiders) because of the auto-indentation on file save. It's back to normal on the VSCode stable version.

ghiscoding:
@jmzagorski 
is this ready for a merge? 
There shouldn't be any errors anymore. Let me know as I'd like to push a release by the end of the week

jmzagorski:
I just rebased these commits and it all looks good. Thanks!

ghiscoding:
Merged... thanks :)

PR #100, opened 2018-09-20
feat(filter): add posibility to filter complex objects

ghiscoding:
In certain cases, our dataset might contain properties that are complex objects. This PR adds the possibility to filter correctly these complex objects.

For example, let say that we have this dataset
```typescript
const dataset = [
 { item: 'HP Desktop', buyer: { id: 1234, address: { street: '123 belleville', zip: 123456 }},
 { item: 'Lenovo Mouse', buyer: { id: 456, address: { street: '456 hollywood blvd', zip: 789123 }}
];
```

We can now filter the zip code from the buyer's address using this filter:
```typescript
this.columnDefinitions = [
  {
    // the zip is a property of a complex object which is under the "buyer" property
    // it will use the "field" property to explode (from "." notation) and find the child value
    id: 'zip', name: 'ZIP', field: 'buyer.address.zip', filterable: true
   // id: 'street',  ...
];
```


ghiscoding:
@jmzagorski 
If you could quickly review this PR, that would be really great. I discovered that I couldn't filter complex objects last weekend and I really want/need this. 

Thanks

ghiscoding:
@jmzagorski 
Yeah at the beginning I had a simple check for the "." notation to kick it in but then I was not sure if all users would be happy to have this kick in at all time. Having the flag was more of an approval that the user does know.... but yeah you have the same thought I had at the start, so I might as well remove it then. Thanks

ghiscoding:
@jmzagorski 
Did the change you requested and after more testing, I found that also the Sort Service needs to be updated. I will do that other change in a future PR, need to wait for this PR to be merged. Could you please review and accept it. Thanks

Also note, I changed the npm `build` script to make it simpler and faster, it will run only the build, without the copy of `dist` to all the other client/demo `node_modules`. The latter can still be done but on a different script now named `build-lib-demo`. This will simplify it on your end.

ghiscoding:
@jmzagorski 
Can you approve the PR after the latest changes? I'd like to merge if possible. Thank you

PR #103, opened 2018-09-21
fix(http): remove aurelia-http-client dependency, closes #101

### LINKED ISSUE #101: Aurelia-Http dependency issue
```
ERROR in ./node_modules/aurelia-slickgrid/dist/native-modules/filters/selectFilter.js
Module not found: Error: Can't resolve 'aurelia-http-client' in '.\node_modules\aurelia-slickgrid\dist\native-modules\filters'
```

* **Library Version:**
2.3.0

* **Operating System:**
Windows 10

* **Bundler used (WebPack/RequireJS/SystemJS)**
Webpack with aurelia cli

* **Browser:**
all

* **Language:**
TypeScript

**Current behavior:**
The library requires the use of aurelia-http-client

The library should not require this library and throw this error

This may not give a good user experience because the grid should work "out of the box". Many user will probably use aurelia's library, but we should not force them to do so, especially since the aurelia CLI does not install the aurelia-http-client with `au new`. 

I found this when starting a new application trying to create a simple 3 column grid where my API call was mocked. This library is found in the select filter and the main custom element used with `instance of` and to define a type for the parameter.

EDIT: Actually I forgot the fetch-client and http-client are two different libraries. Some users may only use the fetch-client.

ghiscoding:
This PR removes the `aurelia-http-client` dependencies. I tested and fixed both Filters & Editors (the latter is in `aurelia-slickgrid.ts`). 

I had to move the `Promise` check as the first check since incorrectly was getting caught by the other checks, after doing so it was all good. I tested with `console.log` on all 3 (Filter & Editor) and switch from `aurelia-http-client`, to `aurelia-fetch-client` to `Promise`.

### DIFF
--- a/aurelia-slickgrid/src/aurelia-slickgrid/aurelia-slickgrid.ts
+++ b/aurelia-slickgrid/src/aurelia-slickgrid/aurelia-slickgrid.ts
@@ -21,7 +21,6 @@ import 'slickgrid/plugins/slick.rowmovemanager';
 import 'slickgrid/plugins/slick.rowselectionmodel';
 
 import { bindable, BindingEngine, bindingMode, Container, Factory, inject } from 'aurelia-framework';
-import { HttpResponseMessage } from 'aurelia-http-client';
 import { EventAggregator, Subscription } from 'aurelia-event-aggregator';
 import { GlobalGridOptions } from './global-grid-options';
 import {
@@ -659,19 +658,19 @@ export class AureliaSlickgridCustomElement {
       // wait for the "collectionAsync", once resolved we will save it into the "collection"
       // the collectionAsync can be of 3 types HttpClient, HttpFetch or a Promise
       //
-      collectionAsync.then((response: HttpResponseMessage | Response | any[]) => {
-        if (response instanceof Response && typeof response.json === 'function') {
+      collectionAsync.then((response: any | any[]) => {
+        if (Array.isArray(response)) {
+          this.updateEditorCollection(column, response); // from Promise
+        } else if (response instanceof Response && typeof response.json === 'function') {
           if (response.bodyUsed) {
             throw new Error('[Aurelia-SlickGrid] The response body passed to collectionAsync was ' +
               'already read. Either pass the dataset from the Response ' +
               'or clone the response first using response.clone()');
           }
-
+          // from aurelia-fetch-client
           (response as Response).json().then(data => this.updateEditorCollection(column, data));
-        } else if (response instanceof HttpResponseMessage) {
-          this.updateEditorCollection(column, response['content']);
-        } else if (Array.isArray(response)) {
-          this.updateEditorCollection(column, response);
+        } else if (response && response['content']) {
+          this.updateEditorCollection(column, response['content']); // from aurelia-http-client
         }
       });
     }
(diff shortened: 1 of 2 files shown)


aurelia-slickgrid/src/aurelia-slickgrid/aurelia-slickgrid.ts
```diff
@@ -659,19 +658,19 @@ export class AureliaSlickgridCustomElement {
...
+        if (Array.isArray(response)) {
+          this.updateEditorCollection(column, response); // from Promise
+        } else if (response instanceof Response && typeof response.json === 'function') {
           if (response.bodyUsed) {
             throw new Error('[Aurelia-SlickGrid] The response body passed to collectionAsync was ' +
               'already read. Either pass the dataset from the Response ' +
               'or clone the response first using response.clone()');
           }
-
+          // from aurelia-fetch-client
```
jmzagorski: might be confusing to have this comment here. A Promise may or may not come from aurelia-fetch-client

aurelia-slickgrid/src/aurelia-slickgrid/filters/selectFilter.ts
```diff
@@ -1,414 +1,413 @@
...
+      // wait for the "collectionAsync", once resolved we will save it into the "collection"
+      const response: any | any[] = await collectionAsync;
 
-            if (!Array.isArray(awaitedCollection)) {
-                throw new Error('Something went wrong while trying to pull the collection from the "collectionAsync" call');
-            }
+      if (Array.isArray(response)) {
+        awaitedCollection = response; // from Promise
+      } else if (response instanceof Response && typeof response['json'] === 'function') {
+        awaitedCollection = await response['json'](); // from aurelia-fetch-client
```
jmzagorski: Same as other comment, Response does not need to come from aurelia-fetch-client

aurelia-slickgrid/src/aurelia-slickgrid/aurelia-slickgrid.ts
ghiscoding: I've put the comments to know where each came from. It was a bit confusing to know the source of each. `aurelia-http-client` and `aurelia-fetch-client`, are not using the same and regular `Promise` are catch by the 1st `if`. Do you have a better way of identifying this?

jmzagorski: The `Response` interface is not specific to `aurelia-fetch-client`. For example, your consumer could be using the native `fetch` api and you would still get the `Response` interface. Not sure you need to identify which API the user is using. I think that is part of the beauty of getting the `Response` object back.

ghiscoding: @jmzagorski 
are you suggesting to remove the comment completely? I just find it a little less clear doing so, but if that removes confusion then sure

jmzagorski: you can keep it if it seems more clear, I have no problem with that. However, be aware that I can use the https://github.com/github/fetch library and it will return `Response` too

ghiscoding: I'll change it for `// from Fetch` then :)

ghiscoding:
@jmzagorski 
Updated the comment and I hope the indentation is back to normal, it should 2 spaces. I will stop using the insiders version of VSCode, that seems to be the cause.

Are you ok to approve and merge?

PR #106, opened 2018-10-03
feat(collection): support multiple filterBy/sortBy in CollectionService

ghiscoding:
- this allows to pass an array or a simple object to both `collectionFilterBy` and `collectionSortBy`

Tested with the following column definition
```typescript
{ 
  id: 'duration', name: 'Duration (days)', field: 'duration', sortable: true, 
  filterable: true,
  filter: {
    collectionAsync: this.http.get<{ option: string; value: string; }[]>(URL_SAMPLE_COLLECTION_DATA),

    // you can pass single/multiple filterBy option(s)
    collectionFilterBy: [{
      property: 'value',
      operator: OperatorType.notEqual,
      value: 365
    }, {
      property: 'value',
      operator: OperatorType.notEqual,
      value: 360
    }],

   // when using multiple filters, it will by default chain the results after each pass
   // this mean that each pass affect all the next passes (less and less results after each passes)
   // but you have the option to change that and use "merge" instead
   collectionOptions: {
     // options are "merge" or "chain" (defaults to "chain")
     filterResultAfterEachPass: 'merge' 
   },

    // you can pass single/multiple sortBy option(s)
    collectionSortBy: [{
      property: 'suffix',
      sortDesc: true,
      fieldType: FieldType.string
    }, {
      property: 'value',
      sortDesc: true,
      fieldType: FieldType.number
    }],
    customStructure: {
      value: 'value',
      label: 'label',
      labelSuffix: 'suffix',
    },
    model: Filters.multipleSelect,
  }
},
```

The collection tested with has multiple items of the following structure. It was modified with only the 1st character of each "suffix" property to add a different char (like "fays", "eays", ...) and our sort is by "suffix", then by "value". 
```typescript
[
  { "value": 0, "label": 0, "prefix": "Task", "suffix": "day" }, 
  { "value": 1, "label": 1, "prefix": "Task", "suffix": "ay" }, 
  { "value": 2, "label": 2, "prefix": "Task", "suffix": "days" }, 
  { "value": 3, "label": 3, "prefix": "Task", "suffix": "fays" }
  // ...
]
```
If we look at the animated GIF below, we see that it works as expected.


ghiscoding:
@jmzagorski 
This PR adds support to pass multiple (array) `collectionFilterBy` and I think there could be room for improvement but I'm not sure how to implement it. Currently what the new code does is to re-use the new filtered collection after each pass (each pass will mean that the collection gets smaller after each pass), but what if the user actually wants to merge the result of all filters instead of removing like it currently does. 

What would be more comfortable with to support both (merge or extract)? I'm thinking about leaving the `collectionFilterBy` interface alone but instead add an extra flag in `collectionOptions`, something like this
```typescript
collectionFilterBy: [{
  property: 'value',
  operator: OperatorType.equal,
  value: 'John'
}, {
  property: 'value',
  operator: OperatorType.notEqual,
  value: 'Doe'
}],
 collectionOptions: {
   separatorBetweenTextLabels: ' ',
   filterBy: 'merge' // new flag can be "merge" or "exclude"
},
```

I'm not totally sure about the flag options (`merge`, `exclude`), if you have better name please provide. The `collectionOptions` already exists (see it's [interface](https://github.com/ghiscoding/aurelia-slickgrid/blob/master/aurelia-slickgrid/src/aurelia-slickgrid/models/collectionOption.interface.ts)), so it's not something new but I could simply add an extra flag into it. 

Thought?

ghiscoding:
I tried a few things and I think I came up with something decent. We can pass an enum flag named `filterResultAfterEachPass` into the `collectionOptions` and the 2 possible options are `merged` or `chained` (`chained` being the default). 

#### 1st example
```typescript
collectionFilterBy: [{
  property: 'value',
  operator: OperatorType.contains,
  value: 10
}, {
  property: 'value',
  operator: OperatorType.contains,
  value: 20
}],
collectionOptions: {
  // options are "merged" or "chained" (defaults to "chained")
  filterResultAfterEachPass: 'merged' 
},
```
If we have an array of value from 1 to 500 with what was written in previous paragraph, the `merged` would return set of values that contains 10 and/or 20, so possible values are `[10, 20, 100, 101, 102,... , 110, ...210, 220, ...]`

while the `chained` would return nothing because the 1st pass returns only value with 10 while the 2nd pass uses result from 1st pass and since value 20 is nowhere to be found in 10 it returns nothing. 

#### 2nd example
collectionFilterBy: [{
  property: 'value',
  operator: OperatorType.notEqual,
  value: 10
}, {
  property: 'value',
  operator: OperatorType.notEqual,
  value: 20
}],
collectionOptions: {
  // options are "merged" or "chained" (defaults to "chained")
  filterResultAfterEachPass: 'merged' 
},
```
When using `chained`, this example would return 498 values (500 items - 2 values), while the `merged` would return the entire 500 items (because both dataset cancels each others)

PR #110, opened 2018-11-10
feat(plugin): split out all controls/plugins into their own class

ghiscoding:
- [x] 1. split out all Controls/Plugins into their own class (now called Extensions)
- [x] 2. use DI for all these new Extensions
- [x] 3. add new Slick.RowMoveManagerPlugin as an extension with example 
- [x] 4. move all Extensions imports from the core lib to the ExtensionService or individual Extension
       - can we use `require` to import plugin only when required (on demand)
       - this would make Aurelia-Slickgrid footprint smaller, only load plugin on demand (when enabled)
- [x] 5. add `unsubscribe` and `destroy` (`dispose`) functions and make sure they are called when destroying grid
- [x] 6. build & test with Aurelia-CLI (RequireJS)


ghiscoding:
@jmzagorski 
This PR might interest you, especially number 4., this also makes (controls & plugins) much more serviceable since every plugins are now separate. 

Nearly done with the PR, just have to test it with Aurelia-CLI

Issue #17, opened 2018-02-26
Feature request - Support a Datalist/Typeahead filter


jmzagorski:
**I'm submitting a feature request**

* **Browser:**
all 

* **Language:**
TypeScript

**Current behavior:**
Not implemented

* **What is the motivation / use case for changing the behavior?**
Many users prefer the datalist/typeahead feature because it is faster than scrolling through a select list. https://github.com/ghiscoding/aurelia-slickgrid/issues/9 is where the conversation got started, and below is a summary. 

1. The html `datalist` is not supported in Safari and partially support in IE
2. Libraries to consider: [purejs-datalist-polyfill](https://github.com/Fyrd/purejs-datalist-polyfill), [jquery datalist polyfill](https://github.com/CSS-Tricks/Relevant-Dropdowns) and [typeahead](https://twitter.github.io/typeahead.js/). The drawback to this is adding another dependency in this library.

ghiscoding:
Safari is becoming the new IE, in terms of following the standards... just add to say it lol

jmzagorski:
yup, when you said that in the original pull request I started rethinking this as a supported feature. Even as I was typing this issue I was wondering if I:

1. Should let the user add their own custom datalist filter so they can pick their favorite library instead of the library getting another dependency
2. Or just support the plain HTML datalist and specify in WIKI it is currently not supported in safari.

Thoughts?

ghiscoding:
Yeah but the thing is that Safari is too big to ignore, I mean it's probably also Safari on mobile and yeah too big to ignore. On my current App project for work, it will be used by internal users, Sales Rep, External Users and I don't have control over the latter. Also if our internal users and Sales Rep uses an iPad or even an iPhone, then I guess it falls on the Safari club. 

Here's a good reference [Can I use - datalist](https://caniuse.com/#feat=datalist), it's not implemented in Safari/iOS/OperaMini though we shouldn't care about the latter. 

What I always wondered is the following though
- is there any way in Aurelia (probably more of a TypeScript question though) to be able create a Service which requires certain lib but if the user does not use that Service (neither does he have the lib installed) then not include it in the Build?

ghiscoding:
BTW, I wouldn't use Typeahead.js as a Filter, it's too big of a lib to use it for a simple Filter. a Polyfill for datalist is much smaller.

[...]

PR #122, opened 2018-12-31
feat(autocomplete): Editor & Filter autocomplete feature 

ghiscoding:
- you can pass "collection", "collectionAsync" or your own datasource implementation
- uses jQuery UI which is a SlickGrid dependency, so no lib added for this feature

This will close issue #17 

List of things TODO
- [x] get a working demo with remote API
- [x] get a working demo with JSON files (objects or string types)
- [x] add SASS styling for the autocomplete 
- [x] add way to provide an optional `placeholder` to tell user what to do 
- [x] run a Build and fix any TSlint issue found

animated GIF of the Feature
w7eyr7qkvp


ghiscoding:
@jmzagorski 
Mostly all working, however there are a couple things to worth knowing
- it uses jQuery [UI Autocomplete](https://jqueryui.com/autocomplete)
    - that jQuery plugin requires a label/value pair, however I managed to go over that by using `customStructure` and using some collection `map` and it's mostly working. 
    - might need some tweaking because of previous point (label/value pair), I'm not sure if it works with everything. We can tweak when we get there
- the Example 3 has 3 new columns for demo purposes
   - "City of Origin" uses a JSONP with Geobytes remote API, that is doable by passing a custom `filterOptions` (or `editorOptions`) with a JSONP `source`
   - "Country of Origin" uses `collectionAsync` with a JSON file with objects having properties (name, code)
   - "Country of Origin Name" uses `collectionAsync` with a JSON file with list of country names (simple strings)
      - I will eventually remove this column, it was only for testing purposes

Issue #271, opened 2019-12-17
package.json typings path incorrect


jbockle:
hey, I'm trying out the latest version 2.15.1 and the typings path in package.json is incorrect, so vscode/tsc is not loading typings.

ghiscoding:
I have no clue what you're talking about.. care to make a PR?

jbockle:
sure

ghiscoding:
I think I understand what you mean after taking a peak, you're talking about this [line](https://github.com/ghiscoding/aurelia-slickgrid/blob/master/package.json#L19) in `package.json`? 

If so, do you know if it's supposed to be pulled from the `amd` or the `commonjs` folder? I moved the entry of the custom element itself couple months ago and I guess the typing should be pulled from `node_modules\aurelia-slickgrid\dist\amd\custom-elements` if it's with `amd`

jbockle:
yeah after installing 2.15.1 in a project and trying to import a module from aurelia-slickgrid, vscode complains that there is no typings and there is no intellisense available.  I can modify the package.json from node_modules/aurelia-slickgrid/package.json to resolve manually for now, PR will fix future instances of this.

ghiscoding:
Merged your PR #272 and pushed a patch release [2.15.2](https://github.com/ghiscoding/aurelia-slickgrid/tree/v2.15.2) version. Thanks for the fix :)

jbockle:
thanks!

PR #272, opened 2019-12-17
(fix) gulp dependency and typings fix

jbockle:
ref #271


jbockle:
@ghiscoding should I increment the version with this PR or do you plan on doing that separately

ghiscoding:
@jbockle 
I bump the version myself whenever I'm ready to push a version. I'll push a patch version one later in the week

PR #336, opened 2020-05-12
fix(gridService): crud methods should support custom dataset id

ghiscoding:


src/aurelia-slickgrid/services/grid.service.ts
```diff
@@ -435,8 +436,9 @@ export class GridService {
   deleteItem(item: any, options?: GridServiceDeleteOption): number | string {
     options = { ...GridServiceDeleteOptionDefaults, ...options };
 
+    const idPropName = this._gridOptions.datasetIdPropertyName || 'id';
     if (!item || !item.hasOwnProperty('id')) {
```
ghiscoding, line 440: this should be `!item.hasOwnProperty(idPropName)`

PR #344, opened 2020-05-17
feat(i18n): add namespace prefix + separator grid option, closes #338

### LINKED ISSUE #338: How to configure used namespace (non-default) for i18n translations?
How to tell the grid which namespace to use for i18n translations?

This setup works in our project. The translations from aurelia-slickgrid.json are displayed in the grid.
```
return instance.setup({
	attributes: ["t", "i18n"],
	backend: {
		loadPath: "/locales/{{lng}}/{{ns}}.json"
	},
	debug: true,
	defaultNS: "aurelia-slickgrid",
	fallbackLng: "fi",
	lng: "fi",
	ns: [
		"common",
		"aurelia-slickgrid"
	]
});
```

The problem is that we already have utilized a different default namespace (common).
So the setup above is not possible and we have to use the one below:
```
return instance.setup({
	attributes: ["t", "i18n"],
	backend: {
		loadPath: "/locales/{{lng}}/{{ns}}.json"
	},
	debug: true,
	defaultNS: "common",
	fallbackLng: "fi",
	lng: "fi",
	ns: [
		"common",
		"aurelia-slickgrid",
		"other-namespaces..."
	]
});
```

1. So how I configure the grid to utilize "aurelia-slickgrid" namespace?
Didn't find any good candidate from interface GridOption.

Now getting a lot of missing key prints in console and the translations are not displayed.
"i18next::translator: missingKey fi common CONTAINS CONTAINS"

2. Also tried to copy all the translations from aurelia-slickgrid.json to common.json but that didn't help either. Still getting those missing key prints. This should work?

3. Moreover getting those missing key prints still after successfully applying the custom locale approach without enableTranslate in grid options. The grid shouldn't look for translations if they are not in use at all.

| Software          | Version(s) |
| ----------------- | ---------- |
| Aurelia           | 0.6.0
| Aurelia-Slickgrid | 2.17.9
| TypeScript        | 3.8.2
| Operating System  | Win10
| NPM/Node/Yarn     | 6.13.4/13.5.0.

ghiscoding:
- ref #338

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,11 +21,6 @@ jobs:
       - run:
           name: Upload Jest coverage to Codecov
           command: bash <(curl -s https://codecov.io/bash) -t ${CODECOV_TOKEN}
-  cypress:
-    working_directory: ~/aurelia-slickgrid
-    docker:
-      - image: circleci/node:10-browsers
-    steps:
       - run:
           name: Start Dev WebServer
           command: npx nps webpack.server
@@ -56,12 +51,3 @@ jobs:
       - run:
           name: Build Demo Site
           command: NODE_ENV=production npx webpack --progress -p --env.production --env.extractCss
-
-workflows:
-  version: 2
-  build-and-test:
-    jobs:
-      - build
-      - cypress:
-          requires:
-            - build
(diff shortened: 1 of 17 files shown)


ghiscoding:
Even if I add this to a util, it won't remove much code. I'll look at it later.

ghiscoding:
@jani-pernu I moved the code into the `utilities.ts` with unit tests, it should all good now. I don't plan on doing more code change

PR #374, opened 2020-06-21
feat(sorting): header menu clear sort, reset sorting when nothing left

ghiscoding:
- when there is no more column to sort, we could resort by default sort id which would display dataset the way it was when it was first loaded
- ref SO question: https://stackoverflow.com/questions/62489108/angular-slickgrid-column-wise-remove-sort-option-not-working-properly
- also found that Clear Column Sort on a Local Grid was not trigger a "sortChanged" event while it should so that the Grid State is also notified

### DIFF
--- a/src/aurelia-slickgrid/services/sort.service.ts
+++ b/src/aurelia-slickgrid/services/sort.service.ts
@@ -339,7 +339,7 @@ export class SortService {
         this.emitSortChanged(EmitterType.local, sortColumns.map(col => {
           return {
             columnId: col.sortCol && col.sortCol.id || 'id',
-            direction: col.sortAsc ? 'ASC' : 'DESC'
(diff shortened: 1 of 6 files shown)


src/aurelia-slickgrid/services/sort.service.ts
```diff
@@ -302,9 +334,25 @@ export class SortService {
 
       grid.invalidate();
       grid.render();
+
+      if (emitSortChanged) {
+        this.emitSortChanged(EmitterType.local, sortColumns.map(col => {
+          return {
+            columnId: col.sortCol && col.sortCol.id || 'id',
+            direction: col.sortAsc ? 'ASC' : 'DESC'
```
ghiscoding, line 342: change it to 
`direction: col.sortAsc ? SortDirection.ASC : SortDirection.DESC`

Issue #424, opened 2020-09-25
Slot for custom footer/header


zewa666:
Hey there, I'm slowly dipping into parts of the library and so naturally some ideas are coming up :)

What I've seen is that while a custom footer is supported, it essentially takes a bare-bones string with a few modifications.
In my case there might be more that will come into the footer, such as a customized pagination.

So what would you say if the [aurelia-slickgrid custom element](https://github.com/ghiscoding/aurelia-slickgrid/blob/master/src/aurelia-slickgrid/custom-elements/aurelia-slickgrid.html#L35) would expose a named slot for both headers and footers like:

```html
<slot name="slick-header"></slot>
<slot name="slick-footer"></slot>
```

so that the consumer could pump in whatever he likes in those places

```html
<aurelia-slickgrid grid-id="grid2"
                       column-definitions.bind="columnDefinitions2"
                       grid-options.bind="gridOptions2"
                       dataset.bind="dataset2"
                       grid-height="300"
                       grid-width="800">
      <div slot="slick-header">A BIG HEADER</div>
      <div slot="slick-footer">That's my footer</div>
    </aurelia-slickgrid>
```

I'd like to discuss this first with you @ghiscoding but can afterwards provide a PR if you'd like

ghiscoding:
I could reply by saying... I don't know much about `slot` yet since I haven't used them yet but I kinda understand the concept. Would it still produce the Footer by using the `showCustomFooter` flag in the Grid Options? I would like to keep that feature

But yeah if it helps with your use, then yes go ahead and create a PR, I'm open for that. 

However a quick thing to note, I never show the Pagination and Footer at the same time, 1 is supposed to replace the other and so I've put some code to only allow displaying a Footer when Pagination is disabled.

zewa666:
Yeah I've seen that part and it makes perfectly sense not allowing both at the same time. The slot approach is completely different though. In essence you would place a slot in your markup and its content can be defined via the consumer. So essentially just a placeholder. There wouldnt be any config needed. If no content for the slot is provided, its not rendered.

zewa666:
I've created the PR along with a new Example29 page, which should hopefully make clear what I'm after :)

ghiscoding:
The only thing that I see as a potential, and very small, issue is that I have some code to remove the height of the Footer/Pagination  from the auto-resize (if you do use that) and you might have to adjust that manually but that's not a big deal, you might need something like this
```ts
this.gridOptions = {
[...]

PR #425, opened 2020-09-25
feat(slots):adds named slots for header and footer 

zewa666:
this adds named slots for a custom header and footer so that more complex scenarios can be injected as part of the aurelia-slickgrid custom element

related issue https://github.com/ghiscoding/aurelia-slickgrid/issues/424

### DIFF
--- a/src/aurelia-slickgrid/custom-elements/aurelia-slickgrid.html
+++ b/src/aurelia-slickgrid/custom-elements/aurelia-slickgrid.html
@@ -1,6 +1,6 @@
 <template>
   <div id="slickGridContainer-${gridId}" class="gridPane" css="width: ${gridWidth}px">
-    <!-- Header slot if you need to create a complex custom footer -->
+    <!-- Header slot if you need to create a complex custom header -->
     <slot name="slickgrid-header"></slot>
 
     <div id.bind="gridId" class="slickgrid-container" style="width: 100%" css="height: ${gridHeight}px"
--- a/src/aurelia-slickgrid/custom-elements/aurelia-slickgrid.html
+++ b/src/aurelia-slickgrid/custom-elements/aurelia-slickgrid.html
@@ -1,5 +1,8 @@
 <template>
   <div id="slickGridContainer-${gridId}" class="gridPane" css="width: ${gridWidth}px">
+    <!-- Header slot if you need to create a complex custom footer -->
+    <slot name="slickgrid-header"></slot>
+
     <div id.bind="gridId" class="slickgrid-container" style="width: 100%" css="height: ${gridHeight}px"
       focusout.delegate="commitEdit($event.target)">
     </div>
@@ -33,5 +36,8 @@
         ${customFooterOptions.metricTexts.items}
       </div>
     </div>
+
+    <!-- Footer slot if you need to create a complex custom footer -->
+    <slot name="slickgrid-footer"></slot>
   </div>
 </template>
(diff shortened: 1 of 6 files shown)


zewa666:
I'm not sure whether you're ok with the newly added Example or if that behavior should be shown as part of another existing Example. If anything else pops up your mind please let me know

ghiscoding:
I'm ok with a new Example, you even added a Cypress test, that is awesome 😉

ghiscoding:
This works really great, nice to such addition, thanks 😄 
Side note, I fixed a small typo before merging.

Also I did try with the auto-resize and as expected a `bottomPadding` (of about 60) was required for it to resize correctly.

zewa666:
Hey thanks for the quick response and the typo fix. Could you point me to where I should update the Wiki to introduce the feature along with your padding fix?

ghiscoding:
The Wikis are public, I guess you could create a new Wiki for that, I didn't have time to create Footer/Pagination Wiki so I might include some info about them later in the same Wiki.

ghiscoding:
@zewa666 
I released a new version on Friday but only added the slot [Example 29](https://ghiscoding.github.io/aurelia-slickgrid/#/slickgrid/example29) today, it's all good. I tried to add slot to my other lib [slickgrid-universal](https://github.com/ghiscoding/slickgrid-universal) which is plain TS and that didn't quite work, I also tried adding `<template>` to the grid container and/or `<slot>` but that didn't quite work, I guess I'll need to read more about how that works in plain HTML (frameworkless), if you have any suggestions to get it working, then let me know. 

Thanks for the contribution :)

zewa666:
Yep both are standards based but rely on using shadow dom. This article should give more insights https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_templates_and_slots

Frameworks like Aurelia or Stencil allow to use those features even on classic dom so that is why by default you wont see it work.

I've added this on purpose to the Aurelia repo as I thought its a very specific feature. But if you're willing to go the shadowdom route we could perhaps enable it for universal as well. I'd need some pointers though how the structure is made. Does it leverage a webcomponent custom element?

ghiscoding:
My other (slickgrid-universal) is framework agnostic, it's roughly the same code as Aurelia-Slickgrid except that I don't have neither use HTML file and I'm using plain vanilla TS/JS to append DOM element to the DOM (on this [line](https://github.com/ghiscoding/slickgrid-universal/blob/master/packages/vanilla-bundle/src/components/slick-vanilla-grid-bundle.ts#L288) from the [vanilla-bundle package](https://github.com/ghiscoding/slickgrid-universal/tree/master/packages/vanilla-bundle), the implementation is mainly the same as the `aurelia-slickgrid` component.

I was just curious to know if I could use it in plain TS, I tried couple of things with/without `<template>` but that didn't fully work, but that's ok I can read more about the Template/Slot, I find that to be an interesting subject and Salesforce also support Slot (as long as my lib creates an ES6 dist, which it does). It's great that Aurelia has it built-in, I wonder how this will work in Au2 though, aren't they supposed to remove `<template>` completely?

zewa666:
The rfc Implementation for au-slot just finished which now allows to use native slots besides au-slots. Templates are still implicitly there you just dont need to write em.

For your sample, youd have to create a custom element which makes use of shadowdom and then it should work. Yeah give the docs a try and ping me if you need help. Besides, perhaps wanna join the Aurelia discord so conversation gets easier than via GitHub issues/PRs ;)

PR #426, opened 2020-09-26
feat(services): add Toggle Filtering/Sorting & Hide Column methods

ghiscoding:
- The new methods are:
   - `filterService.toggleFilterFunctionality()`
   - `sortService.toggleSortFunctionality()`
   - `gridService.hideColumnByIndex(columnIndex)`
   - `gridService.hideColumn(column)`

PR #439, opened 2020-10-07
feat(options): hideHeaderFilterRow option

zewa666:
It would be great if we could have an option to automatically hide the filter row in the header if filtering is turned on. I wasn't able to find anything and have implemented the new option via the `filterService.toggleHeaderFilterRow()` method which of course isn't nice since you see the animation right after loading.

But hopefully this Draft gives you an idea what would be wished.

### DIFF
--- a/src/aurelia-slickgrid/custom-elements/aurelia-slickgrid.ts
+++ b/src/aurelia-slickgrid/custom-elements/aurelia-slickgrid.ts
@@ -277,6 +277,10 @@ export class AureliaSlickgridCustomElement {
       this.sharedService.hideHeaderRowAfterPageLoad = this._hideHeaderRowAfterPageLoad;
     }
 
+    if (this.gridOptions.hideHeaderFilterRow) {
+      this.filterService.toggleHeaderFilterRow();
+    }
+
     // publish & dispatch certain events
     this.globalEa.publish('onGridCreated', this.grid);
     this.dispatchCustomEvent(`${DEFAULT_AURELIA_EVENT_PREFIX}-on-grid-created`, this.grid);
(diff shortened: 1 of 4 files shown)


ghiscoding:
@zewa666 
> It would be great if we could have an option to automatically hide the filter row in the header if filtering is turned on

~Didn't you mean `turned off` though?~ .. oh wait I just saw your new flag, I think you can try this other flag instead, `showHeaderRow: false`. It's used in this [Example 21](https://ghiscoding.github.io/aurelia-slickgrid/#/slickgrid/example21) to hide the filter bar but show an external filter search input. Sorry about the names that are used across the lib aren't always standardized, sometime we see `showX` but other times `hideX`

On another similar story, I've recently added Toggle Filter/Sort/HideColumn, see PR #426, that PR has some code in Example 16 sample code for that but I forgot to update the GitHub Demo, I'll do that later today but you can see it live in my other lib [Slickgrid-Universal - demo](https://ghiscoding.github.io/slickgrid-universal/#/example07)

zewa666:
Oh thats nice, i thought it would turn off the whole header instead :) yeah i walked through the code but really for whatever reason didnt think to give the proposed one a try. I'll check tomorrow and close this afterwards and instead perhaps contribute a more detailed jsdoc for the property

zewa666:
Btw let me just say your examples are so helpful, managed to figure lots of stuff just by properly looking at them

ghiscoding:
@zewa666 
Yeah I built all these Examples & Wikis for everyone to look into and hopefully get less questions hehe, I do refer to them very often myself so it's a win/win for everyone, oh and there's Cypress tests on many of them too. 

I must say that I like people who do take the time and effort to look into the code, analyze it and provide feedback, that is very helpful, so thanks and happy to help you. 😉 

Also if you find that any of my Wikis are missing something, please go ahead and change/update them, they should be editable by everyone.

zewa666:
I've created the follow up PR to improve the jsdoc for showHeaderRow a bit. https://github.com/ghiscoding/aurelia-slickgrid/pull/440

PR #447, opened 2020-10-27
feat(chore): update all npm packages and project to WebPack 4

ghiscoding:
TODOs
- [x] update all npm packages, and use WebPack 4 (can't go to 5 since it's not fully supported in `aurelia/webpack-plugin` yet)
   - there is 1 small exception with `aurelia-pal-nodejs` version `2.0.0`, it was updated to use latest jsdom version which causes some problems in some unit tests (might do it in separate PR)
- [x] run a full build & test with Aurelia-Slickgrid-Demos
- [x] get rid of duplicate `assets` folder, update `webpack.config.js` and Examples to use all same folder
- [x] get rid of `cypress-mochawesome.js` and reprogram Cypress with Mochawesome the same way it works in Slickgrid-Universal
- [x] get rid of `package-scripts.js` and use only npm scripts
- [x] see if in CircleCI we can do a build and then use `http-server` and run the demo to then run Cypress tests (same as Slickgrid-Universal)
- [x] run a full build and test with WebPack BS3 and BS4 demos
- [x] remove WebPack optimization since it creates too many small files (not needed for demo purposes)


webpack.config.js
```diff
@@ -54,10 +72,7 @@ module.exports = ({ production, server, extractCss, coverage } = {}) => ({
       {
         test: /\.css$/i,
         issuer: [{ not: [{ test: /\.html$/i }] }],
-        use: extractCss ? ExtractTextPlugin.extract({
-          fallback: 'style-loader',
-          use: cssRules
-        }) : ['style-loader', ...cssRules],
+        use: extractCss ? [{ loader: MiniCssExtractPlugin.loader }, 'css-loader'] : ['style-loader', ...cssRules]
```
ghiscoding, line 75: we can remove `extractCss` since we always use it

PR #452, opened 2020-11-02
feat(core): switch to ESLint since TSLint was deprecated

ghiscoding:


tsconfig.json
```diff
@@ -25,7 +25,7 @@
     "dist",
     "jspm_packages",
     "node_modules",
-    "**/*.spec.ts"
```
ghiscoding, line 28: remove this change

.eslintrc
```diff
@@ -0,0 +1,167 @@
...
+  },
+  "globals": {
+    "Slick": true,
+    "Slicker": true,
+    "jQuery": true,
+    "$": true,
+    "_": true
+  },
+  "parserOptions": {
+    "project": "tsconfig.json",
```
ghiscoding, line 19: we can remove the `project` since that slows down the linter for no benefit

.eslintignore
```diff
@@ -0,0 +1,5 @@
+**/*.css
```
ghiscoding, line 1: update list to be same as slickgrid-universal

PR #456, opened 2020-11-10
fix(core): clear dataset when disposing and fix few unsubscribed events to avoid leak

ghiscoding:
This is actually quite a big issue since the dataset was persisted and kept growing exponentially, to put in other words the dataset was never cleared even after changing SPA (Single Page App) page and so it kept growing exponentially.

### DIFF
--- a/src/aurelia-slickgrid/filters/inputMaskFilter.ts
+++ b/src/aurelia-slickgrid/filters/inputMaskFilter.ts
@@ -49,10 +49,10 @@ export class InputMaskFilter extends InputFilter {
     // step 2, create the DOM Element of the filter & initialize it if searchTerm is filled
     this.$filterElm = this.createDomElement(filterTemplate, searchTerm);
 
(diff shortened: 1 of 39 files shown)


src/aurelia-slickgrid/filters/inputMaskFilter.ts
```diff
@@ -49,10 +49,10 @@ export class InputMaskFilter extends InputFilter {
     // step 2, create the DOM Element of the filter & initialize it if searchTerm is filled
     this.$filterElm = this.createDomElement(filterTemplate, searchTerm);
 
-    // step 3, subscribe to the keyup event and run the callback when that happens
+    // step 3, subscribe to the input change event and run the callback when that happens
     // also add/remove "filled" class for styling purposes
 
-    this.$filterElm.on('keyup input change', (e: any) => {
```
ghiscoding, line 55: wrong event, it should be `'keyup input'`

Issue #441, opened 2020-10-09
Type-safe column definitions


zewa666:
I was looking at the [Column interface](https://github.com/ghiscoding/aurelia-slickgrid/blob/master/src/aurelia-slickgrid/models/column.interface.ts#L14) and found a generic used. This though is being used for the `dataContext` argument of `asyncPostRenderer`. Now in my opinion this does not make really lots of sense since if not defined otherwise via `params` dataContext is going to be an object with keys named after field names right?

So in order to achieve type-safety for column fields and the renderer here's what I do:

```typescript
type GridFields = "FieldA" | "FieldB" | "FieldC";

export class App {
    public columnDefinitions: (Column & { field: GridFields })[];

    private defineGrid() {
      this.columnDefinitions = [
        { id: "FieldA", name: "Field A", field: "FieldA" /** THIS BECOMES TYPE SAFE **/},
        { id: "FieldB", name: "Field B", field: "FieldB" /** THIS BECOMES TYPE SAFE **/},
        { id: "FieldC", name: "Field C", field: "FieldC" /** THIS BECOMES TYPE SAFE **/,
          asyncPostRender: (cellNode: JQuery<HTMLElement>, row: number, dataContext: {[key in GridFields]: unknown}, colDef: Column) => {
          if (colDef.params.templateUrl && cellNode.length) {
            this.auUtilService
              .createAureliaViewModelAddToSlot(
                colDef.params.templateUrl,
                { model: dataContext.FieldA /** THIS BECOMES TYPE SAFE **/}, cellNode[0], true
             );
          }
        },
        params: {
          templateUrl: "path/to/my/template",
        },
        },
      ];
    }
}
```

With this sample the `field` prop in the columnDefinitions is typesafe and will scream on providing something different, as well as the `dataContext` for the asyncPostRenderer.

Now if we could initialize the Column interface with a generic used for the GridFields equivalent of above, we could also use the generic for the fields and dataContext as I did above manually.

What do you think? I know this is most likely a breaking change, but I really think the current use of the generic doesn't bring any value to the table, except I'm missing a point here.

ghiscoding:
[...]

PR #466, opened 2020-11-27
BREAKING CHANGE: use Slickgrid-Universal monorepo

ghiscoding:
### Next Major Version `3.x` (WIP)

This will be, for the most part, deleting a lot of code and instead use it from the [Slickgrid-Universal](https://github.com/ghiscoding/slickgrid-universal) lib to have 1 common/core lib. Also that lib was built as a monorepo and with that we'll be able to decouple a few Aurelia-Slickgrid Services (Export to File/Excel, OData, GraphQL). So you will end up with a much smaller bundle since you will be bundling only what you really use (without having for example OData/GraphQL that you might never use). For even more info, you can read a full explanation that was posted under an older release [here](https://github.com/ghiscoding/aurelia-slickgrid/releases/tag/v2.19.1) under the section labeled "What's coming in the Future (much later in the year)"

On the migration side, it should be fairly simple, the only changes will be in regards to the Services that get decoupled, you can take a sneak peak at the changes in the upcoming [Migration Guide](/ghiscoding/aurelia-slickgrid/wiki/Migration-to-3.x)

ETA 🤞 - Wrapping it up for Christmas would be nice 🎁 🎄 🎅

#### Minimum Requirement Changes
- TypeScript `>= 4.1.2`
- Aurelia-I18N `>= 3.0.0`
- Aurelia-CLI `>= 2.0.0`
- Bootstrap `>= 4.5.0` (or any other UI framework like Bulma)

##### _if you want to help, just tell me which number you are planning to provide help with_ 
- [x] 1. replace all Formatters with Slickgrid-Universal
- [x] 2. replace all Grouping Formatters with Slickgrid-Universal
- [x] 3. replace all Aggregators with Slickgrid-Universal
- [x] 4. replace all Editors with Slickgrid-Universal
- [x] 5. replace all Editor Valitors with Slickgrid-Universal
- [x] 6. replace all Filters with Slickgrid-Universal
   - (requires DI) extend `AutoCompleteFilter` and `SelectFilter` so that we can watch collection changes with BindingEngine
   - goal change, I might be able to implement collection watch directly in Slickgrid-Universal
- [x] 7. replace all Filter Conditions with Slickgrid-Universal
- [x] 8. replace all Models with Slickgrid-Universal (enums, interfaces)
- [x] 9. replace all Sort Comparers with Slickgrid-Universal
- [x] 10. replace all Services with Slickgrid-Universal
   - (requires DI) need to extend `FilterService` for the 2x Filters mentioned at step 6) to be extended
   - goal change (same as 6.), I might be able to implement collection watch directly in Slickgrid-Universal
- [x] 11. replace all Extensions with Slickgrid-Universal
   - [x] keep RowDetail Extension since it has specific Aurelia code
- [x] 12. replace all SASS Stylings with Slickgrid-Universal
- [x] 13. replace all Export to File Service with Slickgrid-Universal Export to File monorepo package
- [x] 14. replace all Export to Excel Service with Slickgrid-Universal Export to Excel monorepo package
- [x] 15. replace all Export to OData Service with Slickgrid-Universal OData monorepo package
- [x] 16. replace all Export to GraphQL Service with Slickgrid-Universal GraphQL monorepo package
- [x] 17. some of the events might not work on the Aurelia-Slickgrid custom element because the services are now in separate monorepo packages (`onGridBeforeExportToFile`, `onGridAfterExportToFile`, `onGridBeforeExportToExcel`, `onGridAfterExportToExcel`)
   - the names might be different, it should be `onBeforeExportToTextFile`, ...
- [x] 18. add a Translate Service wrapper of Aurelia-I18N for Slickgrid-Universal to work properly
- [x] 19. add a PubSub Service wrapper of Aurelia-Event-Aggregator for Slickgrid-Universal to work properly
- [x] 20. upgrade to latest Aurelia-I18N version
- [x] 21. add demo to cover PR #442 and close issue #441
- [x] 22. use Bootstrap 4 as a base instead of BS3
- [x] 23. drop internal copy of 3rd party lib [multiple-select.js](https://github.com/ghiscoding/aurelia-slickgrid/tree/master/src/assets/lib/multiple-select) and use [multiple-select-modified](https://github.com/ghiscoding/multiple-select-modified) instead
- [x] 24. add `yarn.lock` file into Git and make use of [dependabot](https://dependabot.com/), see [dependabot.yml](https://github.com/ghiscoding/slickgrid-universal/blob/master/.github/dependabot.yml) config from Slickgrid-Universal 
- [x] 25. create/update [Migration Guide](/ghiscoding/aurelia-slickgrid/wiki/Migration-to-3.x) Wiki to next version `3.x`
- [ ] 26. update all necessary Wikis
   - Export to Excel from `@slickgrid-universal/excel-export`
   - Export to Text File from `@slickgrid-universal/text-export`
   - OData Backend Service from `@slickgrid-universal/odata`
   - GraphQL Backend Service from `@slickgrid-universal/graphql`
- [x] 27. rename/move `master` branch into `version2` branch to keep ref to older version
- [x] 28. create a Git Tag on Aurelia-Slickgrid-Demos of `2.x` version before rewriting to newer version
- [ ] 29. add the new SlickGrid Composite Editor + Example
- [x] 30. Convert dynamic loading of SlickGrid plugins/controls from `require(plugin)` to ~`await import(plugin)`~
    - this is causing way too many racing conditions, I will take a look at that much later in the future in separate version.
    - see this [article](https://mariusschulz.com/blog/dynamic-import-expressions-in-typescript) or this other [article](https://blog.mgechev.com/2019/05/11/dynamic-imports-javascript/)
- [x] 31. jQueryUI is imported separately in Slickgrid-Universal (via `main.ts`), check to see if we need to do that here
- [x] 32. search for any leftover of `TODO`, `@ts-ignore` and/or `@deprecated` code
- [x] 33. all Cypress E2E tests should work
- [x] 34. all Jest Unit Tests should work and try to keep a 100% test coverage
- [x] 35. remove a few of the npm scripts, we probably don't need to bundle anymore for (`jspm`, `system`)
- [x] 36. rename monorepo `FileExportService` to `TextExportService` and also couple of option changes:
   - deprecate `exportOptions` use new option `exportTextOptions`
   - deprecate `enableExport` use new option `enableTextExport`
- [x] 37. remove prefix `asg` and `sg` from all events in the global grid options (user can put it back in their own configs).


src/aurelia-slickgrid/services/universalTranslate.service.ts
```diff
@@ -0,0 +1,39 @@
...
+  eventName = 'onLanguageChange' as TranslateServiceEventName;
+
+  constructor(private i18n: I18N) { }
+
+  getCurrentLanguage(): string {
+    return this.i18n.getLocale();
+  }
+
+  async use(newLang: string): Promise<Locales> {
+    return this.i18n.setLocale(newLang);
```
zewa666, line 29: what was the idea with this @ghiscoding? setLocale will return the translation function. I'm not sure I get the intention of this code part. What are those `Locale` supposed to be?

src/aurelia-slickgrid/services/universalTranslate.service.ts
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  setup(options: TranslateOptions) {
+    this.i18n.setup(options);
```
zewa666, line 33: translate options contains lang (guess you meant `lng`) and `loadPath` but this is a configuration of the backend and can't be passed that way. So I'm not sure whether this is a compatibility related introduction or just an oversight.

zewa666:
I've meanwhile pushed the updated example2 and updated to the latest version of aurelia-i18next. Tbh I haven't used i18next yet in side of another library/plugin so I guess the approach going with a peerDependency should be the safest. Will have to check out afterwards with a new project depending on this version.

zewa666:
also with regards to your mentioned BindingEngine issue from Aurelia's discourse. Could you point me to the location where it causes issues and the steps to reproduce? Are you seeing the errors in your unit tests?

src/aurelia-slickgrid/services/universalTranslate.service.ts
ghiscoding: The `Locale` is when the user wants to ignore I18N and define his own Custom Locale, so basically it's when they use a Single Locale, then they don't need to use I18N. English Locale is defined as the default, but you could create a custom one like German Locale and load that Locale (`german.ts`) and they're done. Here's the [Wiki](https://github.com/ghiscoding/aurelia-slickgrid/wiki/Localization-with-Custom-Locales), I know the name might be confusing, I just didn't know what other name to use for representing a locale loading without using I18N.

zewa666: Ok that means you'd like to return the loaded json translation there? Like an object with keynames being locales and values the loaded resources?

src/aurelia-slickgrid/services/universalTranslate.service.ts
ghiscoding: ah yeah I actually didn't look at the full content of that file, I added that in Slickgrid-Universal long time ago and copied over that same file. I'll give another look at it later and reply back, if you think we don't need it then it could be removed. The main point is that the `UniversalTranslateService` wrapper is the most important thing so that both lib works with their own translater service and so far I have achieve that goal, I just forgot about the options part :)

ghiscoding:
I tried extending some of the SlickGrid Filters from the monorepo and it wasn't working at all so I removed the code/files in order to push my PR without making the build fail. My plan was to have the commit building without the files and I'll slowly add back some of those files that need to be extended, I'll do that tonight after work.

src/aurelia-slickgrid/services/universalTranslate.service.ts
ghiscoding: oh wait sorry I didn't notice you commented in the Translate wrapper file, I thought you were in another file. I think you are right, this is wrong and shouldn't return a `Locale` (since that is what I described in previous comment and doesn't fit with I18N). If you have the correct Type from I18N please change it

 Here is how I use it in my other lib
image

and yes you can see I simply load a json file with `require(x.json)`, nothing fancy and I know it might just work with WebPack but it does work with it.

src/aurelia-slickgrid/services/universalTranslate.service.ts
ghiscoding: I've put a print screen of how I use this setup, I'm assuming you can replace the `TranslateOptions` with the correct I18N Type of I18N `.setup()`, I wrote this wrapper in a hurry and thought I was done when I saw it started working but obviously didn't test everything, I haven't tried this setup options yet.

src/aurelia-slickgrid/services/universalTranslate.service.ts
zewa666: Sure I'm still not sure what the returned value of use should be since the type Locales isnt really self-explanatory. As asked above an object keynames as Locales and their value the loaded json or the path to those? Perhaps you can give me an example of how the returned object should ideally look like and I can find you the proper i18next API for it

src/aurelia-slickgrid/services/universalTranslate.service.ts
zewa666: Alright then I'd give it the same type as i18n.setup has.

src/aurelia-slickgrid/services/universalTranslate.service.ts
ghiscoding: It can, and probably should, be the same as what you declared in Au-I18N, which seems to be ` setLocale(locale: string): Promise<i18next.TFunction>` or just put it as `Promise<any>` (what you had in previous Au-I18N). I rarely care or use the return variable anyway, but I do care that it's returning a Promise since I very often want to `await` it (for spinner or text in UI).

For example in 1 of the Aurelia-Slickgrid [demo 12](https://ghiscoding.github.io/aurelia-slickgrid/#/slickgrid/example12), I just await the `setLocale` to display the new locale in the UI next to the switch language button (see below).
image
image

and inside my Slickgrid-Universal interface, I made it return `Promise<any> | any` (I think the `any` was for Angular and I expect Aurelia to return a Promise). However I'm not sure that the description I wrote for the return type is valid, it's kinda of a mix of the Au-I18N and Angular ngx-translate.

So to answer your question, I'm indifferent on the return, so long as it's a Promise with possibly something inside, making it return the same as Au-I18N would be perfect. 

image

zewa666: No question about the Promise I guessed you'd need to await it so clear on that. What does your angular version with ngx-translate return for the same function? Casting any and returning the Translation function from aurelia-i18n is definitely the simplest solution but kinda beats the purpose of a unified API. Since we dont care I'd really like it to do similar to others e.g angular

ghiscoding: So if I remember correctly it returns the translation set of the associated locale (which is pretty much what I wrote in my `use()` interface) and just to be sure I looked at ngx-translate code and that is exactly it as shown on this [line](https://github.com/ngx-translate/core/blob/master/projects/ngx-translate/core/src/lib/translate.service.ts#L203). Now you see where I copied the `use()` method name from, mostly because I wanted the shortest name :P 

Here's a small print screen of this [line](https://github.com/ngx-translate/core/blob/master/projects/ngx-translate/core/src/lib/translate.service.ts#L203) of code, there's a but more code because I think you can override it, but for the most part that is it.

image

ghiscoding: Hey so I made a few commits and wanted the build to pass so I went ahead and made few changes in this file. What I found out was that the `setup(options)` is completely unnecessary, it existed because I copied this file over from Slickgrid-Universal which is a full implementation (or rather very basic) of a Translate Service like I18N is, so that method must exist in the monorepo only for demo purposes and is not needed at all here, so I removed it. I also went ahead and fixed a few of the other comments you've put, if you can take another quick look at it that'd be great (I'm done with the file, so you change whatever else in that file if you wish). Thanks again for the review, your feedback helps **a lot**, really appreciate

zewa666: Sure no prob. Alright thx for the pointers, I'm not sure but perhaps I can get to it today midday when my little boy sleeps.

ghiscoding: I always wonder what TZ you actually are :P ...I'm in Montreal Canada (East TZ), I think we have time diff but I always wonder how many :P

zewa666: Hehe yeah I'm from Austria, GMT+1 so I think 6-7 hours diff. Jeez you talking about Montreal makes me envy. Definitely want to visit that great place again.

zewa666:
Hmm I'm getting a funny error messages of not being allowed to push things to this protected branch due to the codecov issues. do you experience the same?

src/aurelia-slickgrid/services/universalTranslate.service.ts
zewa666: Ok looking at this in detail now I think your approach is right. I'd keep it as promise<any> and the return type as is, since there is not really lots of benefit returning the same nextLang that you gave the function.

ghiscoding:
I released a new version (2.x) yesterday and that caused a few conflicts here and I had a bit of trouble merging and resolving these conflicts. You might need to force push? Unless it's caused by the branch security I added yesterday, but technically it's just for merging when build passes but anyway I removed it just now. Give it another shot plz

zewa666:
Yeah guess that branch security od causing it, can check later at the PC. All official Aurelia packages still ship with all combinations. The reason is to support a wider range of bundlers/loaders like requirejs(amd), rollup(most likeley consuming native) and browserify(commonjs). Since those are autogenerated I would keep them arround if they aren't causing issues for you.

ghiscoding:
ah I didn't know RequireJS uses `amd` so I better keep both `amd` and `native-module`, I'll delete just `system` then. Thanks for the info.

ghiscoding:
@zewa666 
So I worked really hard on this for the entire week because I wanted to get this out the door before Xmas and here's what is left to do:
6. SlickGrid Filters to be extended so that we can use Aurelia Binding Engine watch array changes 
   - for example, the `Filters.MultipleSelect` (and single) have a `collection` property that we can enable a watch which will repopulate the dropdown
10. Filter Service to be extended so that we can use Filter Factory and get step 6 to work
29. Add new SlickGrid Composite Editor modal window component

For the number 6 and 10 (we can say it needs DI), I did a bit of research and I think that I can actually implement that kind of array watch directly in Slickgrid-Universal without having to rely on Aurelia BindingEngine. From what I found, I believe I would be able to do that with a Proxy object and override getters/setters, something like this [SO](https://stackoverflow.com/a/5100420/1212166). I think this can wait and be implemented later on, we rarely have to push new dropdown options after the fact. Also that was the main question that I asked on Discourse which Bigopon replied... I have also manually created my own Services instances and it works but of course isn't ideal, I also added all necessary unit tests for all of that, but we can revisit Bigonpon's solution as well later (if you want to take a look at that, feel free to do so and push changes to the PR).

As for the number 29, that is mainly a big new feature, which I've told you about few weeks ago, and is demonstrated under this Slickgrid-Universal [Example 12](https://ghiscoding.github.io/slickgrid-universal/#/example12) which is like a CRUD modal window with a bit more features... that can wait, I have to take my vanilla approach and create an Aurelia Custom Element for it with all appropriate unit test for it, so it might take a week or two to do so.

So I just released an Alpha version [3.0.0-alpha](https://www.npmjs.com/package/aurelia-slickgrid/v/3.0.0-alpha), I also started to modified 1 of the Aurelia-Slickgrid-Demos (BS4) in this [PR](https://github.com/ghiscoding/aurelia-slickgrid-demos/pull/4), it's a bit more changes than I taught so it might scare a few people, but it is now ready for testing just make sure to read the [Migration Guide](https://github.com/ghiscoding/aurelia-slickgrid/wiki/Migration-to-3.x) that I started writing, it should include most of the changes. 

Give it a shot, let me know
Thanks again

ghiscoding:
@zewa666 
Just a quick update and some questions. I got a way to get the Composite Editor Modal Window working by extracting the component into a separate Lerna package and I will do that in a separate PR because it's a new feature anyway.

So I'd like to merge this PR but I was curious if you had a chance to give a try and/or if you wanted to do more code change/review?

It's looking good for a full release happening quite soon. 
Cheers

zewa666:
Wasnt yet able to take it to a deep dive. I can try to do so tomorrow. Specifically I wanted to check out that no issues with i18n exist.

ghiscoding:
@zewa666 
I updated all the WebPack Aurelia-Slickgrid-Demos and they all work but I didn't have the same luck with the aurelia-slickgrid-demo [cli-requirejs](https://github.com/ghiscoding/aurelia-slickgrid-demos/tree/master/cli-requirejs-demo) bundle, you can see the changes in this [commit](https://github.com/ghiscoding/aurelia-slickgrid-demos/pull/4/commits/85e63b49bea73a3c788c81761387bbe312dd89fd) for it. I get a few errors and the strange thing though is that even if undo the `package.json` and re-run npm install, I get the same kind of errors so could that be just on my side!? I only work with WebPack so not sure what else to do.

I get a bunch of these errors
```bash
Error: File not found with singular glob: C:\GitHub\aurelia-slickgrid-demos\cli-requirejs-demo\node_modules\@slickgrid-universal\excel-export\index.js (if this was purposeful, use `allowEmpty` option)
    at Glob.<anonymous> (C:\GitHub\aurelia-slickgrid-demos\cli-requirejs-demo\node_modules\glob-stream\readable.js:84:17)
```

Things to note, all the Slickgrid-Universal packages are export as 2 dist bundles (commonJS and es2015) and I don't really want to add more dist targets unless I really have to. I add more couple days ago but then when I started extracting the Composite Editor, I ran into "out of memory with CircleCI", so I hope it works with just commonJS. It works perfectly with WebPack, I didn't have time to start a new project with the CLI.

zewa666:
I wonder why it's looking for a file `index.js` in the root of the excel-export folder as it's indicated to use the commonjs one. Very interesting. I'll take a look at that as the sample is the same setup as my app.

zewa666:
@ghiscoding alright this was quick. So good news is it works, bad news is it's ugly.

Seems the sub-dependencies `excel-export`, `graphql`, `odata`, `text-export` and `common` all get the same kind of issue looking for the `index.js` in the root of their node_modules folders instead of `dist/commonjs/index.js`.

The workaround for now is to add
```
{
            "name": "@slickgrid-universal/excel-export",
            "path": "../node_modules/@slickgrid-universal/excel-export/dist/commonjs",
            "main": "index"
          },
          {
            "name": "@slickgrid-universal/graphql",
            "path": "../node_modules/@slickgrid-universal/graphql/dist/commonjs",
            "main": "index"
          },
          {
            "name": "@slickgrid-universal/odata",
            "path": "../node_modules/@slickgrid-universal/odata/dist/commonjs",
            "main": "index"
          },
          {
            "name": "@slickgrid-universal/text-export",
            "path": "../node_modules/@slickgrid-universal/text-export/dist/commonjs",
            "main": "index"
          },
          {
            "name": "@slickgrid-universal/common",
            "path": "../node_modules/@slickgrid-universal/common/dist/commonjs",
            "main": "index"
          }
```

So as said while this works, it's not fixing the root-cause. But at least I can now tell that the new alpha builds properly for requirejs and i18n seems to work as expected. So all clear for my app 🥳

zewa666:
Btw. I'm still not able to push to the branch (the mentioned minor ui fix) due to the same protection issue.

Anyways I'm attaching the patch here as it's really small

```patch
diff --git a/src/examples/slickgrid/example12.html b/src/examples/slickgrid/example12.html
index 4f610bd8..4219619e 100644
--- a/src/examples/slickgrid/example12.html
+++ b/src/examples/slickgrid/example12.html
@@ -4,13 +4,13 @@
 
   <hr />
 
-  <div class="row">
+  <div class="row align-items-center">
     <button class="btn btn-outline-secondary btn-sm" data-test="language-button" click.delegate="switchLanguage()">
       <i class="fa fa-language"></i>
       Switch Language
     </button>
-    <label>Locale:</label>
-    <span style="font-style: italic" data-test="selected-locale">
+    <label style="margin: 0 5px">Locale:</label>
+    <span style="font-style: italic; width: 70px;" data-test="selected-locale">
       ${selectedLanguage + '.json'}
     </span>
```

package.json
```diff
@@ -49,71 +47,54 @@
...
     "slickgrid": "^2.4.32",
-    "text-encoding-utf-8": "^1.0.2"
+    "text-encoding-utf-8": "^1.0.2",
+    "typescript": ">=4.1.2"
+  },
+  "peerDependencies": {
+    "i18next": ">= 14.1.1"
   },
   "devDependencies": {
+    "@slickgrid-universal/excel-export": "^0.4.1",
```
zewa666, line 82: I've noticed the example uses v `0.5.1`. Besides that why is this listed as a devDependency here but as normal dependency for the consumer -> the samples?

ghiscoding: I've pushed `0.5.1` yesterday that includes the Composite Editor as a new Lerna package and was waiting to merge this PR and then create another PR to update to `0.5.1` and add Composite Editor. They are all `devDependencies` because they are really optional, I add them here so I can run Cypress tests and make sure everything works. The only Lerna that is essential is the `/common`, all other ones are optional, does that make sense?

ghiscoding: I added a note in the aurelia-slickgrid-demo saying that those are installed only for demo purposes and should be removed from your dependencies if you don't use the feature, see this [readme note](https://github.com/ghiscoding/aurelia-slickgrid-demos/pull/4/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5)

zewa666: Ok cool. I was wondering whether that was causing the rjs issue

ghiscoding: I assume this is resolve then? Let me know if there's anything else to do before merging this PR, probably later tonight, so I can create the next PR for Composite Editor. Also note that I create a `version2` branch last week, so I still have previous code is something goes awfully wrong.

zewa666: Yeah, as said rjs has only limited spread and there is a workaround. So this shouldnt block the release

ghiscoding:
@zewa666 
Merged in master branch, feel free to create PR for anything else you wanna change. 
Thanks for the great collaboration on this big PR

zewa666:
Big thanks to you for the great effort in unifying also the Aurelia version.

PR #481, opened 2020-12-13
feat(components): add Composite Editor Modal Window component

ghiscoding:
- provide a modal window to Create, Update, Mass Update, Update Selection (selected rows)

image


src/examples/slickgrid/example30.html
```diff
@@ -0,0 +1,67 @@
+<template>
+  <h2>${title}</h2>
+  <div class="subtitle" innerhtml.bind="subTitle"></div>
+
+  <div class="mb-2">
+    <div class="btn-group btn-group-sm" role="group" aria-label="Basic example">
+      <button type="button" class="btn btn-outline-secondary" data-test="toggle-readonly-btn"
+              click.delegate="toggleGridEditReadonly()">
+        <i class="fa fa-table"></i> Toggle Edit/Readonly Grid
```
zewa666, line 9: Since I see it here, are you using an older Version of fontawesome? The newer one requires `fas fa-*`

test/cypress/plugins/utilities.js
```diff
@@ -0,0 +1,18 @@
...
+
+  // then invdate will be 07:00 in Toronto
+  // and the diff is 5 hours
+  var diff = date.getTime() - invdate.getTime();
+
+  // so 12:00 in Toronto is 17:00 UTC
+  return new Date(date.getTime() + diff);
+};
+
+export function zeroPadding(input) {
```
zewa666, line 15: Fear of using left-pad 🤣

src/examples/slickgrid/example30.html
ghiscoding: Yes I've used the old version `4.7.0`, I never tried the new one, I find the new font too fat and the thin one is on the paid side 😆

ghiscoding: if you want to contribute and change it to latest Font-Awesome (or Fort Awesome like I think I saw you using), then I wouldn't mind another PR 😉

test/cypress/plugins/utilities.js
ghiscoding: not sure if you're asking about the function name or the new string padding function which I've never tried? 😝 
it seems I could use this instead `str.padStart(2, '0');`

zewa666: My joke did not catch fire. If you remember the world wide npm issue because of npm left-pad, thats what I was referring to ;)

But yeah other than that padStart certainly od better. Less code = less bugs

src/examples/slickgrid/example30.html
zewa666: Tbh me neither, but since this not the examples repo I think it doesn't matter. Next time around at the samples I'll bump and change to fortawesome.

zewa666:
This is a pretty helpful feature as I find myself often building simple modals to achieve the same. How customizable is this going to be?

ghiscoding:
What kind of customization were you looking for? 

To be fair, it's not extremely customizable because it's auto-generated by reading (looping) through the grid column definitions and you must have editors else it won't show up. When I first demoed this to the business, they liked it however they wanted to do a bit more and that is to disable some field(s) and pre-populate data when certain field are filled (e.g. clicking a checkbox will enable and autofill a date). I've built this just few months ago, so far we are only using it for Mass Update and Selection Mass Changes, we aren't using it for Create/Update yet.. and they might want a Clone a Row feature which would be easy to add, user could clone a row and modify some value then click "clone" and that would create an extra row in the grid. 

So at this point I'm able to
1. disable some field(s) which in term can also disable other field(s)
   - by using the `onBeforeEditCell` and returning False (uses same as logic as inline editing)
2. populate certain field(s) controlled by other field(s)
   - we can add data to the field and/or blank the field (not the same logic as inline editing but similar)
3. validation / custom validation
   - same validation defined in the column definitions but shown directly in the modal
4. it's also responsive (basic)
   - I've added some very basic and generic responsive css classes to make it work in any UI framework
   - it has up to 3 grid containers in the modal and it will use 1, 2 or 3 depending on how many form fields you have and also available space but again it's very basic responsive implementation
   - you can also set to always show 1, 2 or 3 containers in the modal
   - in the example if you click Edit, you'll see it in 2 containers but if you click Mass Update then it will show in 1 because there is 1 less form field and that's the threshold

You can see most of these points in the current Slickgrid-Universal [Example 12](https://ghiscoding.github.io/slickgrid-universal/#/example12) which is the same grid I copied over here in this PR. The only thing I'm not showing is step 2. because I wanted to show validation message (if you click on "Completed", it enables the "Finish" field and says that you need to enter a date, I could have filled it myself programmatically if I want to but I wanted to show validation instead).

Some other things to note is that you have access to all the modal form editors which is how you can control how to populate/change/disabled other fields in the form. Once you click Save, you have access to the form object that was sent and you can do something on frontend/backend.

ghiscoding:
Actually I forgot that I had put that logic in there but if you try the demo and you put the slider editor of "% Complete" to 100%, it will change "Completed" to checked and also fills the "Finish" date to today's date. 

Just curious, did you have other customizations in mind? 

rZQ2JaH5Li

zewa666:
Nope more than happy with what I see. Great work

ghiscoding:
Hey there was another question I had before releasing this, I was at first thinking of using dynamic imports but then I found out yesterday that WebPack, and most probably other bundlers, won't actually tree shake dynamic imports because it doesn't know in advance when it will be loaded. So I resulted back to regular imports assuming that Au2 will have better tree shaking. 

**But** I would really prefer to take full advantage of the monorepo concept and keep a prod build as small as possible and avoid relying on tree shaking to get smaller bundle. What I mean is that I'm trying to find an approach that would assume that you will bundle only what you really want and if the user never uses the Composite Editor then I don't want them to first wait for the import of the Composite Editor inside Aurelia-Slickgrid and then run the build including that piece of code and hoping that the tree shaking process will get rid of the unused code. So I was trying to find better alternative and came up with a different approach (below I'm showing 2 approaches, 1st is the current code in the PR and the 2nd is the new POC).

Let's start with the code in current PR, it is quite simple, you just enable the feature and Aurelia-Slickgrid will register the component when it sees that flag enabled. That is fine but again it doesn't take full advantage of the monorepo and it's yet another 800 lines of code (SlickCompositeEditor) to import which the user might not need. So the current PR's code is the following

#### Approach 1 
```diff
<aurelia-slickgrid grid-id="grid1"
    column-definitions.bind="columnDefinitions"
    grid-options.bind="gridOptions"
+    on-composite-editor-created.delegate="handleOnCompositeEditorCreated($event.detail)">
</aurelia-slickgrid>
```
ViewModel
```diff
export class MyDemo {
  compositeEditorInstance: SlickCompositeEditorComponent;

  initializeGrid() {
    this.columnDefinitions = [ /*...*/];

    this.gridOptions = {
+      enableCompositeEditor: true,
    }
  }

+  handleOnCompositeEditorCreated(compositeEditor: SlickCompositeEditorComponent) {
+    this.compositeEditorInstance = compositeEditor;
+  }
```

That is working fine but again it doesn't take full advantage of the monorepo concept and so I worked on a POC today and I got it working, it's reusing portion of the code that I've done for the Export Services and right now we import these services via the monorepo and we register them like this
```ts
initializeGrid() {
  this.columnDefinitions = [ /*...*/];

  this.gridOptions = {
    registerExternalServices: [new ExcelExportService(), new TextExportService()],
  }
}
```
So I was thinking to use the same concept but change the `registerExternalServices` to something more generic that will represent both External Components & Services, so I think `registerExternalResources` would be a good name. Now for the code, I no longer need the `onCompositeEditorCreated` event to get the instance, instead I create it directly in the example constructor.

#### Approach 2
```diff
<aurelia-slickgrid grid-id="grid1"
    column-definitions.bind="columnDefinitions"
    grid-options.bind="gridOptions"
-    on-composite-editor-created.delegate="handleOnCompositeEditorCreated($event.detail)">
</aurelia-slickgrid>
```
ViewModel 
```diff
import { ExcelExportService } from '@slickgrid-universal/excel-export';
+ import { SlickCompositeEditorComponent } from '@slickgrid-universal/composite-editor-component';

constructor() {
+  this.compositeEditorInstance = new SlickCompositeEditorComponent();
  this.initializeGrid();
}

initializeGrid() {
  this.columnDefinitions = [ /*...*/];

  this.gridOptions = {
-    registerExternalServices: [new ExcelExportService()],
+   registerExternalResources: [new ExcelExportService(), this.compositeEditorInstance],
  }

  openModal() {
+    this.compositeEditorInstance?.openDetails({ /*...*/ });
}
```
With this approach it takes full advantage of the monorepo and I don't think it is much more complicate for the user to code and use. 

Internally though, it's a bit more complicate since the SlickCompositeEditorComponent requires access to some Services so what I've done is to implement a very basic DI Container as `UniversalContainerService` and inside it has 2 methods `registerInstance` and `get` which is similar to Aurelia DI Container, I then manually register each Services (in Aurelia-Slickgrid constructor) and then I loop through all `registerExternalResources` and I call an `init(grid: SlickGrid, container: UniversalContainer)` then inside the SlickCompositeEditorComponent `init` I retrieve all Services I need with `this.container.get('GridService')` and I'm good to go, it all works and it takes full advantage of the monorepo (user will only download/instantiate what he really needs). I made the POC changes in this [commit](https://github.com/ghiscoding/slickgrid-universal/pull/196/commits/6f5a6d8b6c04023352c8f82ba1a294c503e5e040) on Slickgrid-Universal and I ran the TS build and then tried it in Aurelia-Slickgrid and I see it working in both framework. I'm pretty sure that this new `UniversalContainerService` that I created could eventually be replaced by Aurelia DI once I understand how to code what Bigopon mentioned on Discourse and that can be  done after releasing Aurelia-Slickgrid `3.x`. 

So my question to you is, does that make any sense? 
Is that an acceptable new approach? 
...or is that too complicate?

_I know the basic container service might be weird to look at but I'm trying to code something that is framework agnostic and is extendable._

From my point of view, I think the new approach (2) is better because not every user will want the new Composite Editor and if they do then it's the same as registering the ExcelExport (or TextExport) and so they're pretty much all aligned now. As for the `UniversalContainerService`, I can always refactor that in pure Aurelia DI later on and if I do then it should be transparent to the users of my lib. Also with this approach, I could always add more Services/Components and use this coding style to add more features without affecting the core lib itself and without affecting the user's code either. 

Sorry for the long post, I just want to make sure that I'm on the correct path 😉

zewa666:
I cant speak fully for Webpacks dynamic import tree shaking as I dont have enough experience with Webpack itself. If it would work though it would result in an easier API right? Perhaps @bigopon can shed some light on this?

ghiscoding:
hmmm I'm still thinking the 2nd approach is preferable because it aligns with the approach I've used for the Export Service and I'm sure that I only import what I really want and only importing when reaching that page (delayed import). Anyway, I need to make a decision in coming days because I need to make sure that Slickgrid-Universal still works in our Salesforce project and Thursday is my last working day this year (I tried this 2nd approach and I'm still ok).

zewa666:
There is nothing wrong with the 2nd approach from a quick look. So once we know whats the issue with dynamic it might get into the next breaking changes.

ghiscoding:
I dropped dynamic imports after reading couple of articles, [Dynamic Imports and Tree Shaking](https://medium.com/better-programming/dynamic-import-and-tree-shaking-in-javascript-ddc2f3cd69f) and [WebPack Modules](https://apiumhub.com/tech-blog-barcelona/webpack-modules/)

The first article at some point says
> Essentially, dynamic import is known as a non-tree-shaking supportive feature.
...
> Why Dynamic Import Doesn’t Support Tree Shaking
> You may wonder why this feature doesn’t support tree shaking. It’s because of the unclear importing timing.

That last article also makes it quite clear
> Webpack can not know in advance if the dynamic import is going to be used at some point, so in those cases you can not apply Tree Shaking and can not be sure that dynamic imports will never be used.

So from my understanding it's a common issue for all bundlers, not just WebPack and that is why I looked for other approaches to take advantage of the monorepo structure.

ghiscoding:
@zewa666 
So I decided to go with the 2nd approach and I just tested something I had in mind with Aurelia, so I created a very basic Container Service in Slickgrid-Universal and I intended to possibly just extend it in Aurelia-Slickgrid by using the Aurelia-Framework Container if possible.... and to my surprise it just works :)  Basically I call a bunch of `container.registerInstance('MyService', this.service)` (because I instantiate all services manually and outside of DI) and from within Slickgrid-Universal it calls the basic container with `container.get('MyService')` and it's auto-magic :)  

I might be able to get rid of those registration completely later on, that is if I can create the bridge that Bigopon proprosed on discourse, but for now I'm happy with the result. 

Note that what I just described is only on my local, I will push another commit tomorrow, I need to do yet another patch version of Slickgrid-Universal to take full advantage of this. Getting closer and closer...

zewa666:
Sounds promising indeed. Looking forward to it

PR #492, opened 2020-12-25
feat(ci): replace CircleCI by GitHub Actions

ghiscoding:


zewa666:
That looks pretty ... slick ;)

The only thing I'm unsure is the npm Install together with using yarn. Isnt that conflicting? Also with npm you'd go with `npm ci` to speed up installs and make them use fixed versions. I dont use yarn so is there an equivalent?

zewa666:
Ahh ok so thats what bahmutov/npm-install@v1 does under the hood. Cool lgtm

ghiscoding:
ahh right I used Yarn in Slickgrid-Universal because of Lerna with Yarn Workspaces but I guess I don't need it here though. I could remove it. The `bahmutov/npm-install@v1` is what I found to work with Cypress being inside another sub-folder (not in the root) as it was describe in Cypress Github Action [here](https://github.com/cypress-io/github-action#subfolders)

zewa666:
Yep just read that as well. Perhaps wait for the switch until you switch to npm7 workspaces and make sure to check in the package-lock.json as well. bahmutov/npm-install@v1 will handle that as well

zewa666:
I also thought since aurelia-slickgrid and universal is Open source you should be able to request the [Cypress OSS](https://www.cypress.io/oss-plan/) plan to make use of the Dashboard as indicated by @bahmutov s action you can directly push artifacts there

ghiscoding:
hmm I tried adding the recording key the other day and it seems that I reached their limit quite fast, but now the link you sent seems to show no recording limit for Open Source project, perhaps I didn't do it correctly or they changed it again.

zewa666:
Ah and while on it, perhaps running the lint task would be a good idea as well

ghiscoding:
the lint is already on the build npm script as a prebuild ;)

ghiscoding:
guess I'm done... anything else before I merge? :P

zewa666:
Lgtm, :shipit:

PR #524, opened 2021-02-14
fix(backend): incorrect item count with GraphQL and useLocalFiltering

ghiscoding:
- [x] requires latest version of Slickgrid-Universal


src/examples/slickgrid/example25.ts
```diff
@@ -222,11 +248,11 @@ export class Example25 {
    * We also need to resolve the data in a flat array (singleSelect/multipleSelect Filters only accept data at the root of the array)
    */
   getLanguages() {
```
ghiscoding, line 250: the code should be updated to include interface
```ts
getContinents() {
    const continentQuery = `query { continents { code, name  }}`;
    return this.http.post<GraphqlResult<Country>>(COUNTRIES_API, { query: continentQuery });
  }
```

and 
```ts
getCountries(query: string): Observable<GraphqlResult<Country>> {
    return this.http.post<GraphqlResult<Country>>(COUNTRIES_API, { query });
  }
```

PR #567, opened 2021-05-01
fix(tree): few issues and huge improvement found in Tree Data

ghiscoding:
1. initial sort not always working
2. tree level property should not be required while providing a parentId relation
3. tree data was loading and rendering the grid more than once (at least 1x time before the sort and another time after the tree was built) while it should only be rendered once

#### TODOs
- [x] requires Slickgrid-Universal [PR #326](https://github.com/ghiscoding/slickgrid-universal/pull/326) to be merged and released
- [x] should work with all 3 things mentioned on top
   1. initial sort not always working
   2. tree level property should not be required while providing a `parentId` relation
   3. tree data was loading and rendering the grid more than once (at least 1x time before the sort and another time after the tree was built) while it should only be rendered once
- [x] add missing unit tests
- [x] need to investigate, it seems to remove the tree level property when original dataset doesn't have the property (e,g, `indent`)
- [x] do more tests in our SF implementation
- [x] add throw error to not use multi-column sorting with Tree Data
- [x] should also work when adding new item dynamically as shown in the demo 
  - there was an issue in the demo because the demo also has a filter preset
- [x] use `gridService.addItem(x)` (and `addItems()`) and make sure that it refreshes & resort the hierarchical dataset by itself (user shouldn't have to do anything except calling this service addItem)
- [x] it should also work with `updateFilters` and `updateSingleFilter`
- [x] add throw error when using backend service api with Tree Data since it isn't supported
- [x] export to file/excel should also export the tree and its indentation
- [x] refactor flat to tree structure and vice versa to optimize speed
   - so far it seems to be ~40-50x faster 😮🚀 (previous code to convert from tree to flat was extremely slow, deep copy and item search were the main culprit). 
   - with 100k rows test after sorting "Title" column
      -  (before) 3354.64892578125 ms  
      -  (after) 70.06689453125 ms
- [x] add ways to change dataset dynamically in the demo (2 buttons to change from 500 to 25k rows)
- [x] throw error when trying to use Tree Data without filtering enabled (this is required for Tree Data)
- [x] we shouldn't sort the dataset if user didn't provide the `initialSort` property
   - when not provided that means it's assuming the user already sorted the dataset, possibly while querying it from the dataset. However this is not recommended to do, we still have that option, technically speaking Example 5 is already sorted because of how it is constructed).
- [x] remove previous throw error of multi-columns sort, it seems to work after all (need to test a bit more)
  - [x] it seems to be broken after all, so we should probably leave it as unsupported
- [x] rename some of the methods, the word "hierarchical" can simply be renamed to "tree"
- [x] update all unit tests
- [x] try adding more Cypress E2E tests to cover at least insert of items
   - [x] test that clear filters shows all original items count
- [x] remove all benchmark console logs before merging the PR
- [x] add optional prefixes to the Tree Formatter rows
- [x] test that we should still be able to add new item in both Example 5 and 6
- [x] add pub/sub events (`onBeforeSortChange`, `onBeforeFilterChange`, `onBeforeCollapse`, ...) to help display spinner when dealing
- [x] adding new item should also update formatter on Title column but it doesn't currently do that
- [x] add SASS vertical alignment variable for the tree collapsing icons
- [x] add custom title formatter in tree data options for the tree formatter to process


src/examples/slickgrid/example27.html
```diff
@@ -11,25 +11,42 @@ <h2>
...
 
+  <div class="row" style="margin-bottom: 4px;">
+    <div class="col-md-12">
+      <button class="btn btn-outline-secondary btn-sm" data-test="add-500-rows-btn" click.delegate="loadData(500)">
+        500 rows
+      </button>
+      <button class="btn btn-outline-secondary btn-sm" data-test="add-50k-rows-btn" click.delegate="loadData(25000)">
+        25k rows
+      </button>
+      <button onclick.delegate="dynamicallyChangeFilter()" class="btn btn-outline-secondary btn-sm"
```
ghiscoding, line 22: oops there's a typo here it should be `click.delegate`

PR #617, opened 2021-07-23
fix(aurelia-slickgrid): reference html explicitly

MaximBalaganskiy:
By default, aurelia webpack plugin won't look for html in a mono repository `node_modules` folder.
This change will allow for usage in mono repositories without additional webpack configuration.

### DIFF
--- a/src/aurelia-slickgrid/custom-elements/aurelia-slickgrid.ts
+++ b/src/aurelia-slickgrid/custom-elements/aurelia-slickgrid.ts
@@ -77,7 +77,7 @@ import {
 import { SlickFooterComponent } from '@slickgrid-universal/custom-footer-component';
 import { SlickEmptyWarningComponent } from '@slickgrid-universal/empty-warning-component';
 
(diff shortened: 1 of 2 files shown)


ghiscoding:
hey thanks for the contribution, do you mind looking at the Jest unit tests they seem to fail. Also does your change help in VSCode or something like it? If you could add a print screen of this, that'd be great. Thanks

MaximBalaganskiy:
Duh, forgot to import... This won't affect VS Code in any way.
`AureliaPlugin` normally searches for `html` files only in project's `node_modules`. If a project is a part of a monorepository then the root `node_modules` wouldn't be scanned unless `viewsFor` parameter is supplied to the plugin.
With explicit `html` reference this addidional webpack config is not needed.

ghiscoding:
@MaximBalaganskiy hey thanks again for the contribution I made a quick patch version `3.9.2` before officially starting my vacation, check the [Release](https://github.com/ghiscoding/aurelia-slickgrid/releases) for the complete picture. Cheers ⭐

PR #647, opened 2021-09-05
feat(backend): rollback on error & add cancellable events

ghiscoding:
- the lib will now rollback Sorting, Filtering or Pagination if an error is thrown by a Backend Service API (OData, GraphQL). User can optionally cancel the rollback of search input values by disabling `resetFilterSearchValueAfterOnBeforeCancellation`
- add 3x new cancellable events `onBeforeSort`, `onBeforeSearchChange`, `onBeforePaginationChange`
- for the events to be cancellable, I had to replace the local PubSub Service (which was using Aurelia Event Aggregator) by a custom EventPubSub (which uses JS dispatch custom event that are cancellable)

#### TODOs
- [x] requires newer version of Slickgrid-Universal and SlickGrid

PR #649, opened 2021-09-06
feat(tree): add `dynamicallyToggledItemState` method to toggle parent(s)

ghiscoding:
- [x] requires newer version of Slickgrid-Universal
- [x] also requires PR #647 to be merged prior to this PR


test/cypress/integration/example27.spec.js
```diff
@@ -5,7 +5,7 @@ function removeExtraSpaces(text) {
 }
 
 describe('Example 27 - Tree Data (from a Hierarchical Dataset)', { retries: 1 }, () => {
```
ghiscoding, line 7: change the text to "from a flat dataset with parentId references"

PR #656, opened 2021-09-28
feat(core): use Slickgrid-Universal Pagination reusable component

ghiscoding:
- for a lighter lib size and for reusability, we'll reuse the Slickgrid-Universal Pagination across all frameworks


src/aurelia-slickgrid/custom-elements/aurelia-slickgrid.ts
```diff
@@ -544,6 +546,7 @@ export class AureliaSlickgridCustomElement {
 
     // dispose the Components
```
ghiscoding, line 547: actually missing the dispose for the empty warning
```ts
this.slickEmptyWarning?.dispose();
```

ghiscoding: seem to be missing aurelia-slickgrid unit test changes as well on the footer/pagination side of things

PR #1038, opened 2023-08-08
feat!: upgrade to Aurelia 2, closes #709

### LINKED ISSUE #709: Aurelia 2 support
## I'm submitting a Feature request
Are there plans to upgrade to Aurelia 2?

MaximBalaganskiy:
As discussed, copied the upgraded files keeping the original location.
I was a bit confused by the demo app structure, so I've modified it just enough that it compiles.
Non-converted TS files got "ts_old" extension so that compiler does not pick them.

One thing that got me during conversion is that in many examples the mock data method mutates the array AFTER it's already bound to an element. This results in weird grid issues. So the way to do it was to build an array first and then assign to the field.


package.json
```diff
@@ -1,163 +1,96 @@
...
+    "build": "rimraf dist && webpack --env production",
+    "analyze": "rimraf dist && webpack --env production --analyze",
+    "test": "jest",
+    "prepare": "npm run build"
   },
-  "aurelia": {
-    "build": {
-      "resources": [
-        "aurelia-slickgrid/aurelia-slickgrid"
+  "jest": {
```
ghiscoding, line 69: could you put back the original stuff on top of the file (npm scripts, repo info and others...) and also this `jest` section is in a separate folder of its own so not required to be in here since that will fail my own unit tests.

src/aurelia-slickgrid/custom-elements/aurelia-slickgrid.ts
```diff
@@ -495,6 +500,12 @@ export class AureliaSlickgridCustomElement {
     this._eventPubSubService.publish(`${aureliaEventPrefix}onAureliaGridCreated`, aureliaElementInstance);
   }
 
+  /** This is an Aurlia lifecycle hook */
+  detaching() {
+    this.detached();
```
ghiscoding, line 505: is `detaching` a new lifecycle hooks that replaces `detached`? If so then we might as well just renamed `detached` to `detaching`, or is that something different?

src/aurelia-slickgrid/custom-elements/aurelia-slickgrid.ts
```diff
@@ -160,50 +164,51 @@ export class AureliaSlickgridCustomElement {
...
-      groupingAndColspanService?: GroupingAndColspanService,
-      paginationService?: PaginationService,
-      resizerService?: ResizerService,
-      rxjs?: RxJsFacade,
-      sharedService?: SharedService,
-      sortService?: SortService,
-      treeDataService?: TreeDataService,
-    }
+    // TODO: MB - not sure what this is for
+    // externalServices: {
```
ghiscoding, line 174: this `externalServices`, as mentioned in our discussion, was used only for my Jest Unit Tests since I was instantiating the custom element directly (basically without DI, it's easier for me to test when instantiating manually) and in Angular-Slickgrid I use `@Inject('externalServices')`, does Au2 have anything similar to inject something custom without necessarily passing the usual DI? Instantiating my own arguments also allows me to pass mock services more easily than using the typical Aurelia testing structure which I had problem with in Au1 at least.

src/aurelia-slickgrid/extensions/slickRowDetailView.ts
```diff
@@ -275,11 +268,10 @@ export class SlickRowDetailView extends UniversalSlickRowDetailView {
   // ------------------
 
   protected disposeViewSlot(expandedView: CreatedView) {
-    if (expandedView && expandedView.view && expandedView.viewSlot && expandedView.view.unbind && expandedView.viewSlot.remove) {
+    if (expandedView && expandedView.controller) {
```
ghiscoding, line 271: ah it looks like I missed it earlier but we could use optional chaining here `?.`, it's also nice to see that it got way simpler in Au2 for dynamic components

src/aurelia-slickgrid/models/aureliaViewOutput.interface.ts
```diff
@@ -1,7 +1,8 @@
...
+import { ICustomElementController } from '@aurelia/runtime-html';
 
 export interface AureliaViewOutput {
-  bindings?: any;
-  view?: View;
-  viewSlot?: ViewSlot;
+  // bindings?: any;
+  // view?: View;
+  // viewSlot?: ViewSlot;
+  controller?: ICustomElementController<any>;
```
ghiscoding, line 7: so all these 3 props got replaced by 1 new `controller` prop and that's it? I will have to keep that in mind for the migration since that will be a breaking change for sure. I also assume you left the comments for me to see the differences, you could go ahead and delete them now that I've seen it :)

src/aurelia-slickgrid/services/aureliaUtil.service.ts
```diff
@@ -1,73 +1,22 @@
...
+    @IAurelia private readonly au: IAurelia
   ) { }
 
-  createAureliaViewModelAddToSlot(templateUrl: string, bindableData: any, targetElement?: HTMLElement | Element, clearTargetContent = false): AureliaViewOutput | null {
-    const viewFactory = this.viewCompiler.compile('<template><compose view-model.bind="template" view-model.ref="viewModelRef"></compose></template>', this.viewResources);
-
+  async createAureliaViewModelAddToSlot(viewModel: Constructable, bindableData: ViewModelBindableInputData, targetElement?: HTMLElement | Element, clearTargetContent = false): Promise<AureliaViewOutput | null> {
     if (targetElement) {
+      // TODO: MB - is this needed?
       if (clearTargetContent && targetElement.innerHTML) {
```
ghiscoding, line 13: yeah if I remember correctly it removes any html from the RowDetail (or what is behind the opened row detailed) and avoids showing html over html (or row detail on top of slickgrid cells). I think it was for that but I could be wrong

src/examples/slickgrid/example19.ts
```diff
@@ -33,7 +33,7 @@ export class Example19 {
   // extensions!: ExtensionList<any>;
   flashAlertType = 'info';
   message = '';
-  subscriptions: Subscription[] = [];
+  // subscriptions: Subscription[] = [];
```
ghiscoding, line 36: it doesn't look like it's being using so might as well delete it

tsconfig.json
```diff
@@ -1,43 +1,20 @@
...
-    "lib": [
-      "es2018",
-      "dom"
-    ],
-    "types": [ "i18next", "i18next-xhr-backend" ],
-    "typeRoots": [
-      "node_modules/@types",
-      "src/typings"
-    ]
+    "target": "ES2017",
```
ghiscoding, line 8: can we target higher than 2017? I was targeting ES2018 before, we can probably go with at least ES2018 and perhaps even ES2020 for Au2, what do you think?

ghiscoding (changes requested):

ghiscoding:
@MaximBalaganskiy great thanks a lot for this PR, I've done a quick review without running the code yet. You could also give me access to write on your branch so that I could maybe fix some stuff myself within the same PR if possible (or I could merge and fix them later since it's on the `next` branch anyway). 

Side note, we're working (mainly myself) on the next SlickGrid (core lib) as well and there's still more work to do on that project, I've converted that old JS code project to TypeScript and ES6/ESM (see our [SlickGrid Roadmap](https://github.com/6pac/SlickGrid/discussions/807) and I'll probably align the releases at the same time as the Au2 support since there are some breaking changes in the core lib as well (we'll no long have the SlickGrid on the `Slick` window object, we'll be using regular imports instead and that is a change in some of the examples, for example the Example 3 uses `Slick.GlobalEditor` that will have to be replaced by  `import { SlickGlobalEditor } from 'slickgrid'` (or `from 'aurelia-slickgrid'`, I'm not sure yet).... anyway long story short, I assume there's still work until Sept (back to school lol) on that side of it as well.

https://github.com/ghiscoding/aurelia-slickgrid/blob/85cee527c9d824c4f8b114208ff91ea70a6f1d07/src/examples/slickgrid/example3.ts#L631-L634

src/aurelia-slickgrid/custom-elements/aurelia-slickgrid.ts
MaximBalaganskiy: Yeah, `detaching` is a new hook. I left `detached` there because it's got a parameter which is not compatible with the new signature

src/aurelia-slickgrid/custom-elements/aurelia-slickgrid.ts
MaximBalaganskiy: We can define an interface and inject it similar to `IContainer` for example

src/aurelia-slickgrid/models/aureliaViewOutput.interface.ts
MaximBalaganskiy: I assumed that these were only needed to properly dispose of the dynamic content, so not sure about anyone relying on them

src/aurelia-slickgrid/services/aureliaUtil.service.ts
MaximBalaganskiy: I think the question here is whether you can have situations when you DON'T need to clean the inner html?

tsconfig.json
MaximBalaganskiy: I'll try. From my previous experience, es2020 did not work with Aurelia1 decorators, not sure about AU2

MaximBalaganskiy:
The access should be there. As for the timeframe, I'm personally not in a rush, I've got an Aurelia1 project which is not in production (might not even be released in the end) and upgrading it is not on my list atm.

src/aurelia-slickgrid/custom-elements/aurelia-slickgrid.ts
MaximBalaganskiy: Actually, I would just inject those services and let DI create them. You can provide mocks during testing as before.

src/aurelia-slickgrid/custom-elements/aurelia-slickgrid.ts
```diff
@@ -108,13 +99,26 @@ export class AureliaSlickgridCustomElement {
   private _isLocalGrid = true;
   private _paginationOptions: Pagination | undefined;
   private _registeredResources: ExternalResource[] = [];
+  private _columnDefinitionObserver: ICollectionObserver<CollectionKind.array>;
+  private _columnDefinitionsSubscriber: ICollectionSubscriber = {
+    handleCollectionChange: () => {
```
ghiscoding, line 104: @MaximBalaganskiy 
I'm back to working on this migration and that this `_columnDefinitionObserver` only works once and not after, would you have any clue why that is? 

to replicate the issue, you can get latest code and go to Example 3 then click on the "Dynamically Duplicate Title Column" button multiple times, it only add an extra column on the first time but not after and after debugging and adding a console log in here, I see that the subscriber only worked once

MaximBalaganskiy:
I think replacing the array makes the subscription inactive. The first time
it works because you push first. Either do not replace the array or
resubscribe when it changes.

ghiscoding:
> I think replacing the array makes the subscription inactive. The first time it works because you push first. Either do not replace the array or resubscribe when it changes.
> […](#)
> On Fri, 15 Sep 2023 at 1:46 pm, Ghislain B. ***@***.***> wrote: ***@***.**** commented on this pull request. --------------------

I digged into Aurelia 2 docs and found that the `xChagned` is still supported (as per [docs](https://docs.aurelia.io/components/bindable-properties#calling-a-change-function-when-bindable-is-modified)) and if I add back simple code like this
```ts
columnDefinitionsChanged(newCols, oldCols) {
    console.log('columns changed', newCols, oldCols);
}
```
then the console log does get increase by another column every time I click the button. The only thing that I see from Au2 docs is that compare to Au1 it doesn't call the initial Changed event but we can call it ourselves from the `bound()` method... so was there any reason you decided to go with the approach you had? Was it to make the code a bit cleaner or...? Should I put back previous implementation with `columnDefinitionsChanged`?

MaximBalaganskiy:
It was probably an oversight. I like to keep props and changed handlers
together - different to your code style - that is why I missed the
connection and deleted that method.

ghiscoding:
ok no problem, I'll add it back since it worked better, I just might need to run it once for the initial load to make it the same as Au1 but at least it should make it work again :)

Thanks for the info... and time to go to bed, it's passed midnight here :O

ghiscoding:
@MaximBalaganskiy 
Looking back at the code, it looks like I need both because just one doesn't fulfill all use cases
1. `columnDefinitionsChanged` for assignment like `this.columns = newColumns[...];` or `.slice()`
2. `getArrayObserver`, which used to be `bindingEngine.collectionObserver`, for bindings and array methods like `.push`, `.pop` and others...

I see that you migrate step 2 because of this Au2 migration [BindingEngine has been removed in v2](https://docs.aurelia.io/developer-guides/migrating-to-aurelia-2#binding-engine) comment but with the current `bindingEngine.collectionObserver` with Au1, it was working fine and I only needed to create and subscribed to the observer just once in the `bind` lifecycle and I was done, but it looks like with Au2 the `getArrayObserver` really does need to be re-subscribed like you mentioned earlier since it seems to lose its subscription somehow. Do you know why that is? Do you know if there's something else that I can use which can combine both assignment and array methods observing? 

https://github.com/ghiscoding/aurelia-slickgrid/blob/a842b750e1ac074d1dfea64bbcd3c0b84c4d3069/src/aurelia-slickgrid/custom-elements/aurelia-slickgrid.ts#L581-L598

MaximBalaganskiy:
Not sure why they made it this way. I'm not a big expert in au2 so not
aware of other alternatives. I've found documentation is lacking with
regards to observation, or may be I'm blind :)

ghiscoding:
@MaximBalaganskiy 
I got a lot further with the migration and now have all Examples working and all Cypress E2E tests passing, I'm still missing 90% of Jest unit tests but I can fix that later. The main problem I'm getting now is with the WebPack production build, it's not working as I expect to be. After a WebPack prod build, I like to use [servor](https://www.npmjs.com/package/servor) which is a small http server for making sure the prod build is working as it should. However it doesn't work with this new WebPack config for Aurelia 2. 

At first, I got this error
> [webpack : Uncaught ReferenceError: require is not defined](https://stackoverflow.com/questions/45818937/webpack-uncaught-referenceerror-require-is-not-defined)
which as this Stack Overflow answer suggest, we shouldn't be using `nodeExternals()` in a prod build, and if I comment that out then the build works without error but nothing shows in servor. 

Do you have a working setup for a production build? I'm pretty sure that it has to work with my servor little server, if it doesn't then the GitHub demo page won't work either. Any chance you can help with this? For now, I'll use WebPack in dev mode to run CI E2E tests

MaximBalaganskiy:
The prod build only builds a plugin, what are you serving after it?

ghiscoding:
> The prod build only builds a plugin, what are you serving after it?

@MaximBalaganskiy The build for publishing the lib on npm is a separate process, what I want to fix is the website build for the GitHub demo page and also for the CI

ghiscoding:
> The prod build only builds a plugin, what are you serving after it?

ohhhh I see what you mean now, I was not using WebPack for the plugin build in Au1 but it looks you are in Au2, I did not noticed that. I was using TypeScript build previously, would that still work or should I really need to use WebPack for the plugin build?

Anyway, here is what I modified in WebPack to get the website demo working as I wanted. 

image

As I mentioned in previous paragraph, I was using simple TypeScript scripts to build the plugin, could I still use these scripts or should I really use WebPack to build the plugin?

https://github.com/ghiscoding/aurelia-slickgrid/blob/a3567d6c6d24603cefafa4b61f87c09df4207841/package.json#L34-L41

ghiscoding:
I think it's time to merge since this PR is starting to be quite large already, however the Aurelia-Slickgrid component unit test is entirely missing, I seem to have problem running the unit tests for that one in particular but that can be added in the future (or skipped entirely, it doesn't make much difference on the usage since all Cypress E2E tests are passing).

Please note however that a new major release is still a month or two away since there need to be major restructure in the Slickgrid-Universal project.

MaximBalaganskiy:
No problem, I wasn't expecting such a quick upgrade anyway. Nice to know that it will definitely happen now.
Thanks for your commitment!

ghiscoding:
Hey so I just wanted to give an update on the Aurelia 2 upgrade, I'm still working on the next major version of Slickgrid-Universal and it took me much longer than expected but anyway I'm pretty sure that I'll be done in the next couple weeks (before Christmas for sure), for more info on what is left to, you can take a look at the Slickgrid-Universal [Roadmap 4.0](https://github.com/ghiscoding/slickgrid-universal/discussions/1108). Cheers

PR #1128, opened 2023-12-17
chore: migrate to npm workspaces & tsc builds, fixes #709

### LINKED ISSUE #709: Aurelia 2 support

ghiscoding:


ghiscoding:
@MaximBalaganskiy is not run from the root which is why it's failing, its root is actually in `test` and I was passing the config like this `--config ./test/jest.config.ts`. You can disable the Jest tests for now if it's too much work

MaximBalaganskiy:
@ghiscoding It runs but I get a lot of errors

ghiscoding:
yeah no problem, I can take a look at it later.. however I'm wondering why the Cypress E2E tests aren't running

MaximBalaganskiy:
No idea why prod build is failing, runs fine on my machine

ghiscoding:
@MaximBalaganskiy you can maybe save the dist folder in the CI to inspect it, so that you can download the zip of what's bring used by the CI

```ts
       - name: Archive common build artifacts
         if: ${{ inputs.dryrun != true }}
         uses: actions/upload-artifact@v3
         with:
           name: dist-common
           path: |
             packages/common/dist
           retention-days: 5
```

MaximBalaganskiy:
Can I leave the rest to you? I'm not familiar with Cypress at all

ghiscoding:
yeah thanks a lot for everything my friend, you are definitely the person who makes this Aurelia 2 migration possible 🥇

MaximBalaganskiy:
@ghiscoding row detail seems to work fine in a demo build
image

ghiscoding:
@MaximBalaganskiy 
So I got it all working, the failing tests was due to dynamic components being commented out. One last question though, is there any reason why you used `peerDependencies` instead of `devDependencies` in the `packages/aurelia-slickgrid`?

MaximBalaganskiy:
Peer deps will make sure that all requirements are met

ghiscoding:
right I had my mind somewhere else, however I think peer dep should be ok just for Aurelia 2 itself, users should be required to install all packages of Slickgrid-Universal themselves (especially for `@slickgrid-universal/common`, it should be installed automatically when installing Aurelia-Slickgrid) and I think that should be the same for all other Aurelia deps. I'll just keep Aurelia 2 as peer dep.

MaximBalaganskiy:
@ghiscoding 
I think npm installs peer deps automatically anyway so it doesn't matter

ghiscoding:
for npm yes but not necessarily for other package manager. I'm pretty sure it's find to just have Aurelia 2 in peer dep, that's typically what I've seen so many times in React projects (I'm not a React developer but I see so many React articles and never Aurelia so..)

So anyway, I think I'm all done and ready for another release, finger cross that it all works now... but pretty confident since now we have npm workspaces to test with real dist output. That's a much better approach for testing the project

ghiscoding:
@MaximBalaganskiy 
thanks a lot for your help, v7.0.3 is finally a good release and is working fine with Aurelia 2 (finally)!!! However, after updating the Aurelia-Slickgrid-Demos, I'm facing problems with the Single Locale (without i18n DI), that was working fine in Aurelia 1, do you know if there's a way to make a DI optional in Aurelia 2? For example, in Angular-Slickgrid I can use `@Optional()` to say that a dependency can be optional (so I'm using `@Optional() protected readonly translaterService: TranslaterService`)) but I don't see anything to that regard in Aurelia 2 docs. I did not try to simply make it optional with `?:` but I think that will help since I don't see much of a difference in the dist output

MaximBalaganskiy:
See this https://github.com/aurelia/aurelia/blob/6a79bc9b4fb96a25d62470d83739f89f7fc7135a/packages/kernel/src/di.ts#L618C4-L618C4

ghiscoding:
ok that's helpful, the optional part is working after commenting out the i18n registration, great.. but now the i18n doesn't work, I'm using this now `@optional('TranslaterService') protected readonly translaterService?: TranslaterService` but I'm not sure where to put the injection key to be used and now it basically says that my Translater Service is not loaded in my i18n demo when it should.

MaximBalaganskiy:
Use the class as the key, not a string

ghiscoding:
oh ok, but actually I think I have to make i18n optional instead directly into my Translater Service like this maybe

```ts
export class TranslaterService implements UniversalTranslateService {
  constructor(@optional(I18N) @I18N private readonly i18n: I18N) { }
````

ghiscoding:
@MaximBalaganskiy I've just pushed a new release v7.1.0 which includes the code shown above, so now both with & without I18N work and I got them both working in the [Aurelia-Slickgrid-Demos](https://github.com/ghiscoding/aurelia-slickgrid-demos) repo. Thanks a lot again for all the help you provided, this new release is now fully working with Aurelia 2 (finally) 🚀

MaximBalaganskiy:
Xmas starts early! 🎉

PR #1166, opened 2024-03-07
chore(deps): refresh lock file to fix security issue found in `ip` pkg

ghiscoding:


packages/aurelia-slickgrid/src/services/aureliaUtil.service.ts
```diff
@@ -20,6 +20,6 @@ export class AureliaUtilService {
 
     targetElement.innerHTML = `<${def.name} model.bind="bindableData.model" ${addonBindable} ${gridBindable} ${dataViewBindable} ${parentBindable}></${def.name}>`.trim();
 
-    return { controller: await this.au.enhance({ host: targetElement, component: { bindableData } }) };
+    return { controller: await this.au.enhance({ host: targetElement, component: { bindableData } }) as any };
```
bigopon, line 23: From beta 12, aurelia doesn't return a controller anymore, but an `IAppRoot` wrapper, like it used to. You can do just like this

```suggestion
    return await this.au.enhance({ host: targetElement, component: { bindableData } });
```

If your application is throwing `Router is already started` error, then it's because `.enhance` call is trying to invoke the app tasks again. This is unintentional change in behavior, and can be fixed by
```ts
```suggestion
    return await this.au.enhance({
      host: targetElement,
      component: { bindableData },
      container: this.au.container.createChild().register(AppTask.creating(() => {}))
    });
```
The above code will shadow the app tasks registered in your root application.

ghiscoding: oh yes that last one seems to work nicely, thanks a lot @bigopon. I'm not sure if the error could be improved or not, but it was very confusing to troubleshoot

PR #1348, opened 2025-05-03
feat: update Aurelia 2 Beta 24

ghiscoding:


ghiscoding:
@bigopon from your post in Aurelia discourse

> thanks, so just to be clear for the topic we already have: the use case is to wash away existing content of an element with `.innerHTML = html` and call `.enhance` on that element again and again. I do think it’s a valid case, probably we should have some tweaks to make it support such scenario.

I'm not exactly sure what you mean with regards to my code, can you elaborate?

is it because of this code below? 

https://github.com/ghiscoding/aurelia-slickgrid/blob/1d740b640d4264d4ffcc744c2b1212fe0a3b8b43/packages/aurelia-slickgrid/src/services/aureliaUtil.service.ts#L17-L23

bigopon:
@ghiscoding 
for this part
> ... wash away existing content of an element with .innerHTML = html

I meant this code
https://github.com/ghiscoding/aurelia-slickgrid/blob/1d740b640d4264d4ffcc744c2b1212fe0a3b8b43/packages/aurelia-slickgrid/src/services/aureliaUtil.service.ts#L17-L23

And for
> ... call .enhance on that element again and again

I meant your `.enhance` call

ghiscoding:
@bigopon so are there any code that I can use that would work then? This util was just a helper for the end user to easily create a component

bigopon:
immediately it can be worked around using `delete element['$au']`. I'll be coming up with a better way.

ghiscoding:
@bigopon Yeah I would rather find a workable temp solution until you make another release, so you meant something like this?

```ts
targetElement.innerHTML = `<${def.name} model.bind="bindableData.model" ${addonBindable} ${gridBindable} ${dataViewBindable} ${parentBindable}></${def.name}>`.trim();
delete (targetElement as any)['$au'];

return await this.au.enhance({
  host: targetElement,
  component: { bindableData },
  container: this.au.container.createChild().register(AppTask.creating(() => { }))
});
```
that doesn't seem to help since I'm still getting the same error it seems. There's probably a cache somewhere?

```sh
dom.node.ts:31 Uncaught (in promise) Error: Node already associated with a controller, remove the ref "au:resource:custom-element" first before associating with another controller
    at set (dom.node.ts:31:15)
    at _Controller._hydrate (controller.ts:407:5)
    at app-root.ts:132:20
    at onResolve (functions.ts:433:10)
    at app-root.ts:131:14
    at onResolve (functions.ts:433:10)
    at new AppRoot (app-root.ts:94:28)
    at _Aurelia.enhance (aurelia.ts:77:34)
    at AureliaUtilService.createAureliaViewModelAddToSlot (aureliaUtil.service.ts:20:28)
```

bigopon:
Ahh yes my bad, should be deleting the property inside $au, not the object itself. It would be the `au:resource:custom-element` property

ghiscoding:
Awesome that works and is a good temp solution to upgrade, thanks a lot 🚀 

`delete (targetElement as any)['$au']?.['au:resource:custom-element'];`

Side note, I was a little surprised to see you hijack the DOM element that way, it probably make sense though since that is the element being the target to the enhancing, but still was a surprise to me. Are there any Type helper instead of using `any` for this kind of things though?

ahh reviewing your PR [code](https://github.com/aurelia/aurelia/blob/31ccb37ccfa76f41ece8181092ce344f96549d54/packages/runtime-html/src/dom.ts#L17-L18) and I guess this is what I should use
`delete (targetElement as Writable<INode>)['$au']?.['au:resource:custom-element'];`

